PR #45, opened 2026-07-30
DEBUG: win-gnu stress-heaps backtrace (do not merge)

zackees:
Temporary debugging PR to capture a gdb backtrace for the win-gnu-only `test-stress-heaps` segfault. Not for merge; will be closed once the backtrace is captured.


zackees:
Debugging complete — root cause found and fixed in #44. The win-gnu `test-stress-heaps` failure was a MinGW-specific thread-cleanup leak: the loader TLS callbacks were registered with MSVC-only pragmas that GCC ignores, so `DLL_THREAD_DETACH` never fired and every exiting thread leaked its theap and pages (RSS 34.64 GB -> 0.02 GB, theaps 1857 -> 5). Closing this temporary branch.

Issue #67, opened 2026-07-31
Phase D: zero-tracking (zalloc skips memset after zero-purge)


zackees:
Sub-issue of #61 (Phase D). **At most one Phase D feature is adopted, and this is the one.**

From Bun's fork: track when a purge left a range reading back zero, so `mi_zalloc` can skip its `memset`. Rated **4/5** in `MIMALLOC_FORKS.md` — self-contained, no API surface, lowest interaction risk of the three candidates.

Not adopting (reasons in #60): the background scavenger (adds a thread that races our profiler's per-page state; purge can decommit a page a live sample still references) and hole purging (+1000 lines in the hottest file, makes every future rebase permanently harder).

### Rule
**Any flag that ships must have a CI job running the full suite with it ON. If we cannot afford the job, we cannot afford the flag.** Plus a removal trigger: not on by default within two releases ⇒ reverted.

### Required stress coverage
- Correctness: every zero-purge path must actually yield zeroed memory. A wrong answer here is silent heap corruption, so the test must assert zeroness across size classes, alignments, and after purge/reuse cycles.
- Concurrent with an active profile and dump.
- Interaction with `MI_SECURE`, `MI_PADDING`, `MI_DEBUG_FULL`.
- The anti-workload: a pattern designed to make the feature **lose**.

### Acceptance
Measured win **and** anti-workload number published. **Dropped if it regresses anything** — neutral is the most likely outcome, and neutral features that ship live forever untested.

Refs #61

zackees:
**Measured — results in PR #79.** Summary:

| Platform | sparse (favourable) | **dense (ANTI-WORKLOAD)** | vs noise |
|---|---|---|---|
| **windows** | -96.4% | **-10.8% faster** | **real** (spreads 3-4%) |
| ubuntu | +0.2% | +0.6% | no effect |
| macos | +4.5% | +3.8% | uninformative (spreads 32-67%) |

It clears this issue's bar — a win on the anti-workload, no regression anywhere — but with caveats that must not be dropped: it is **Windows-only in practice** (Linux purges via `MADV_DONTNEED`, which never sets `needs_recommit`, so the guarded branch does not run), and **macOS is unmeasured rather than neutral**.

**Retargeted to v4** per #61 (rev 4). It cannot go on by default until macOS is measured on something quieter than a shared runner, and the two-release removal trigger starts when v4 opens.

zackees:
**Merged into `v4`** via #79 — resolved, not dropped.

[...]

Issue #61, opened 2026-07-31
META: adopt community mimalloc improvements into v3, and prove each one works [COMPLETE -- v4 work continues in #94]


zackees:
Community mimalloc improvements — Bun's fork above all — land on a **new v4 line**, not on v3.

> **Revision 4.** Rev 3 planned to fold community work into v3. Phases A, B and C are now **done and merged**, and doing them changed the conclusion: every remaining community item is either a fork-identity change, a rebase-cost commitment, or gated on moving the upstream pin. None of those belong in the line that is currently published as `0.9.x`. So this issue is re-scoped from *"adopt into v3"* to *"open v4 and adopt there."* Change log at the bottom.

---

## Governing principle (unchanged, and it paid off)

**A change claiming a performance or memory improvement does not merge without a check that would detect its regression.** Correctness fixes are **not** gated on benchmarks — they gate on a test that fails without the fix.

Why any gate: this repo shipped **two unbounded memory leaks** (#44, #47) that the whole suite passed, because every test asked *"did it crash?"* and never *"did memory stay bounded?"*

That principle earned its keep during Phases A–C. Three separate times, the thing that caught the bug was a **positive control**, not review:

- the arm64 instruction scanner matched **nothing** and reported `PASS ... no above-baseline instructions`
- the memory gate's tolerance came from an assumption ("memory is the low-variance signal here") that measurement disproved — real spread is 6–12% per run
- a local A/B "measured" a 13% regression that reversed sign when re-run interleaved

None were C bugs. Two were Python. All three would have shipped as green.

---

## What is already done (v3, merged)

| Phase | Outcome | Issues |
|---|---|---|
| **A** — memory + counter gate | **Live** on 3 platforms with committed per-platform baselines, min-of-4 sampling, and a positive control that must catch an injected leak (verified at +212% / +98% / +27%) | #63 / #70 |
| **B1** — shared-library CI | Green on ubuntu + win-gnu. Found 4 test targets hardcoding `mimalloc-static` that broke shared-only builds, and a real MSVC DLL bug (#69, still open) | #62 / #68 |
| **B2** — `MIMALLOC_PROF*` namespace | **Keep the names.** The prefix falls out of `mi_option_init` hardcoding `mimalloc_`; renaming relocates the collision rather than removing it. `MI_PROF_CONFIG_OVERRIDE` is the real escape hatch and is now tested on every env-backed field | #65 / #73 |
[...]

PR #79, opened 2026-07-31
feat: experimental zero-tracking behind mi_option_purge_zeroes (#67)

zackees:
Refs #67. **Opened to get the measurement, not to claim a result.** See "verdict" below.

After a decommit-purge the range reads back zero on recommit, but the arena's dirty bits stayed set — so the next allocation was reported non-zero and `mi_zalloc` did a redundant `memset`. This clears the dirty bits on that path, **off by default**.

## Two guards, both load-bearing

Claiming zero wrongly hands dirty memory to `mi_zalloc` — silent heap corruption, not a visible failure. So:

- only on the `needs_recommit` branch (a genuine decommit). A **reset** (`MADV_FREE`/`MEM_RESET`) may preserve contents.
- only when the arena has no custom `commit_fun`, which makes no zeroing promise at all.

Both verified by running the correctness test with `MIMALLOC_PURGE_DECOMMITS=0` as well as the default.

The option is appended at the **end** of the `mi_option` enum rather than beside the other purge options, so the numeric values published in 0.9.0 don't shift.

## Correctness

Fills **every byte** of every block with `0xA5` before freeing, then requires the next `mi_zalloc` to return genuinely zeroed memory across purge/reuse cycles.

This matters: my first benchmark checked 3 bytes per block and never dirtied the middle, so it could not distinguish "zeroed" from "stale but happens to be zero" — it would have passed a corrupting implementation. Registered twice in ctest, off and on, per #67's rule that *any flag that ships must have a job running the suite with it ON*.

## No verdict yet — deliberately

Local numbers do not support one, and I nearly published a wrong one:

| attempt | anti-workload result |
|---|---|
| sequential A then B | **+13% slower** — looked decisive |
| same binaries, interleaved | **6 of 8 faster** — sign reversed |

The difference was measurement order on a machine running concurrent builds. So this PR ships the harness rather than a conclusion: the `zero-tracking` workflow runs the arms **interleaved and paired from a single binary** on all three runners, and `ci/zero_tracking_report.py` prints medians with the within-arm spread, explicitly flagging any effect smaller than the noise.

One thing already clear: **the favourable workload is not evidence on its own.** It shows ~−97%, but touching only one byte per block means skipping the memset also skips faulting the pages in. That is lazy commit, not zeroing. The anti-workload decides adoption.

## Merge criteria

Per #67, this merges only on a measured win with the anti-workload number published, and is **dropped if it regresses anything**. If CI shows neutral, the right outcome is to close this unmerged — a neutral feature that ships lives forever untested.

### DIFF
--- a/.github/workflows/zero-tracking.yml
+++ b/.github/workflows/zero-tracking.yml
@@ -1,0 +1,50 @@
+name: zero-tracking
+
+# Decides issue #67: does mi_option_purge_zeroes actually pay?
+#
+# Deliberately its own workflow rather than a step in c-unit: timing is only meaningful
+# on a quiet runner, and an A/B that shares a job with a parallel build measures the
+# build. A first local attempt "measured" a 13% regression that turned out to be
+# measurement order on a loaded machine -- the interleaved rerun reversed the sign. So
+# the arms here are interleaved and paired, from a single binary, and the raw numbers
+# are printed rather than a verdict.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'src/arena.c'
+      - 'test/bench-zero-tracking.c'
+      - '.github/workflows/zero-tracking.yml'
+
+jobs:
+  ab:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - run: cmake -B build -DMI_PPROF=ON -DCMAKE_BUILD_TYPE=Release
+      - run: cmake --build build --config Release
+      # Correctness first: every byte of every mi_zalloc must read zero, with the
+      # feature both off and on. If this fails the timings are irrelevant.
+      - name: Correctness (feature off, then on)
+        shell: bash
+        run: ctest --test-dir build -C Release -R zero-tracking --output-on-failure
+      - name: Interleaved A/B
+        shell: bash
+        run: |
+          set -euo pipefail
+          exe=$(find build -name 'mimalloc-test-zero-tracking*' -type f | head -1)
+          export MIMALLOC_PURGE_DELAY=0
+          for i in 1 2 3 4 5 6 7 8; do
+            MIMALLOC_PURGE_ZEROES=0 "$exe"
+            MIMALLOC_PURGE_ZEROES=1 "$exe"
+          done | tee ab.jsonl
+          python3 ci/zero_tracking_report.py ab.jsonl
+      - uses: actions/upload-artifact@v4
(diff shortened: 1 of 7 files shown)


zackees:
## Measurement complete — 8 interleaved pairs per runner

| Platform | sparse (favourable) | **dense (ANTI-WORKLOAD)** | distinguishable from noise? |
|---|---|---|---|
| **windows** | **-96.4%** | **-10.8% faster** | **Yes** — within-arm spreads 3.6% / 3.0-4.2% |
| ubuntu | +0.2% | +0.6% | No — spreads 1.3-5.3%. No effect. |
| macos | +4.5% | +3.8% | No — spreads 32-67%. Uninformative. |

Correctness passed on all three runners with the option **off and on**, plus with `MIMALLOC_PURGE_DECOMMITS=0` locally to exercise the reset path the guard must reject.

### Reading this honestly

**The win is real on Windows and it is on the anti-workload**, which is the number #67 said would decide adoption. -10.8% with within-arm spreads of 3-4% is well outside the noise, and this is a Windows-first repo.

**It does nothing on Linux.** +0.2% / +0.6% is not a small win, it is *no effect* — almost certainly because Linux purges with `MADV_DONTNEED`, which does not set `needs_recommit`, so the guarded branch never runs. That is the guard working as designed rather than a bug, but it means the feature is Windows-only in practice and the issue text should say so instead of implying a portable win.

**macOS is unmeasured, not neutral.** Spreads of 32-67% on that runner swamp any effect this size. The `+3.8%` sign is consistent across both workloads, which is mildly suspicious, but at that noise level it means nothing. **Do not ship this on by default until macOS is measured on something quieter** — the honest current status is "no evidence either way".

The sparse column is *not* evidence and should not be quoted alone: touching one byte per block means skipping the memset also skips faulting the pages in, so -96% is mostly lazy commit.

### Against #67's acceptance criteria

> Measured win **and** anti-workload number published. **Dropped if it regresses anything.**

Win: yes, on Windows, on the anti-workload. Regression: none observed anywhere. So it clears the bar it was set.

zackees:
**Holding this open as a draft, not merging into `main`.**

Per #61 (rev 4), community improvements now target a **v4 line** rather than v3. This is a Bun-derived feature, and it is Windows-only in practice, so it is exactly the kind of change that line exists for.

Blocked on:
1. #80 — the upstream pin bump, which is the first v4 change and defines the branch
2. #61 Open Question 1 — whether v4 is a new branch or becomes `main`

Retarget and merge once the v4 branch exists. Nothing here needs rework: correctness passes on all three runners with the option off and on, and the measurement is done.

The one thing still outstanding before it could ever go **on by default** is a macOS number from something quieter than a shared runner — 32-67% spread there means the current result is no evidence either way.

Issue #92, opened 2026-07-31
cross: build job uploads a partial dist/ when compilation fails, hiding the cause


zackees:
The `cross` workflow's build job can fail to compile test binaries, **upload a partial `dist/` anyway**, and let the downstream test job fail with no diagnostic.

## What happened

On a PR whose content was byte-identical to a green `main`, two of six targets failed. The test job's entire log for the failing step was:

```
Run generator=$(find dist/x86_64-unknown-linux-gnu -maxdepth 1 -type f -name 't3_stats-*' | head -n 1)
  test -n "$generator"
  MIMALLOC_PPROF_CI_DUMP="$PWD/heap.prof" "$generator"
  ...
##[error]Process completed with exit code 1.
```

No output. `test -n "$generator"` had exited 1 because the binary was not there.

The reason was four directories up, in the **build** job:

```
soldr: an intermediate can be reclaimed while a slow link is still reading it -- see soldr#1969.
soldr: retrying usually succeeds
error: could not compile `mimalloc-pprof` (test "t4_accum") due to 1 previous error
error: could not compile `mimalloc-pprof` (test "t9_stats") due to 1 previous error
error: could not compile `mimalloc-pprof` (test "t10_snapshot_concurrency") due to 1 previous error
error: could not compile `mimalloc-pprof` (lib test) due to 1 previous error
```

A known soldr compile-cache race. The build nevertheless proceeded to `xargs -r -a bins.txt -I{} cp {} "dist/$TARGET/"` and uploaded whatever *had* built.

## Two distinct problems

**1. The build job does not fail when compilation fails.** `xargs -r` treats an empty/short `bins.txt` as success, so a partial build produces a green build job and a poisoned artifact. Any downstream consumer then fails for reasons that have nothing to do with its own logic.

**2. The failure is undiagnosable at the point it surfaces.** `test -n "$generator"` prints nothing. It took reading a different job's log to find the cause, and the shape of the failure (two targets, reproducible across three reruns, identical content passing on `main`) actively suggested a code bug -- I initially and wrongly attributed it to a sampling-determinism bug (#91).

## Ask

1. **Fail the build job if any expected binary is missing.** Assert the count in `bins.txt` matches expectation, or explicitly check for each required test binary before uploading.
2. **Make the artifact self-validating.** If a `dist/` upload can be partial, the consumer should say *what* was missing: replace `test -n "$generator"` with a message naming the target and listing what *was* staged.
[...]

Issue #91, opened 2026-07-31
mi_prof_start_seeded is not deterministic: the TLD address is mixed into the seed


zackees:
`mi_prof_start_seeded(rate, seed)` and `MIMALLOC_PROF_SEED` are documented as giving reproducible sampling. They do not.

## The bug

`src/profile.c`, `prof_random`:

```c
static uint64_t prof_random(mi_profiler_tld_t* tld) {
  uint64_t x = tld->random;
  if (x == 0) x = prof_seed ^ (uintptr_t)tld ^ UINT64_C(0x9E3779B97F4A7C15);
  ...
```

The per-thread PRNG is seeded from `prof_seed` **XOR the address of the TLD**. That address varies run to run under ASLR, so the same seed produces a different sample sequence every time. The seed narrows the space; it does not fix it.

The README currently claims:

> `MIMALLOC_PROF_SEED=N` -- Deterministic sampling, for repeatable tests

which is not true as written.

## Why the address is there, and why the fix is not just deleting it

Mixing in `tld` is doing real work: without it, every thread starting from the same `prof_seed` would have an **identical** sample sequence, so N threads would sample at the same byte offsets instead of independently. Removing the term would trade non-determinism for correlated threads, which is worse.

The fix is to mix in something that is per-thread but **deterministic** -- a monotonically-assigned thread ordinal, or a counter incremented under the existing `prof_lock` at TLD init -- so thread K always gets `hash(prof_seed, K)` regardless of where its TLD happened to land.

## How this surfaced

`cross` is intermittently red on the pprof round-trip step, on the same code:

| Run | Branch | Result |
|---|---|---|
| 22:36 | `feat/zero-tracking-probe` | **failure** |
| 22:40 | `feat/zero-tracking-probe` (rerun) | success |
| 23:08 | `main` | success |
| 23:17 | `ci/v4-triggers` (identical content to main) | **failure**, reproducible on rerun |

Failing targets were `x86_64-unknown-linux-gnu` and `x86_64-pc-windows-msvc`; `aarch64-unknown-linux-gnu`, `x86_64-pc-windows-gnu`, and both darwin targets passed in the same run.

`rust/mimalloc-pprof/tests/t3_stats.rs` asserts `(100..=400).contains(&objects)` after `common::start(524_288, 43)` -- a fixed seed that is not actually fixed. A run whose sample count falls outside that window fails, and the CI step that greps the emitted profile fails with it.

[...]

PR #93, opened 2026-07-31
ci(cross): fail the build when compilation fails, instead of shipping a partial dist/

zackees:
Fixes the diagnosis half of #92, and unblocks #89.

## What happened

A PR whose content was **byte-identical to a green `main`** failed on 2 of 6 targets, reproducibly across three reruns. The entire log for the failing step:

```
Run generator=$(find dist/x86_64-unknown-linux-gnu -maxdepth 1 -type f -name 't3_stats-*' | head -n 1)
  test -n "$generator"
##[error]Process completed with exit code 1.
```

No output — `test -n` prints none. The cause was four directories away, in the **build** job:

```
soldr: an intermediate can be reclaimed while a slow link is still reading it -- see soldr#1969.
error: could not compile `mimalloc-pprof` (test "t4_accum") due to 1 previous error
error: could not compile `mimalloc-pprof` (test "t9_stats") due to 1 previous error
...
```

The build reported **success** and uploaded a partial `dist/`.

## Two defects

**1. pipefail is off.** GitHub's default shell is `bash -e` *without* `pipefail`, so

```sh
eval "$BUILD_COMMAND ... --message-format=json" | jq ... > bins.txt
test -s bins.txt
```

reports **jq's** exit status. A failed cargo build looks green, and `test -s` passes because *some* binaries built. Now `set -euo pipefail`.

**2. Nothing verified the artifact was complete.** The downstream job needs `t3_stats` specifically; that is now asserted in the build step, next to the compiler output, and it lists what *was* staged rather than failing silently.

## Worth recording

I initially misattributed this to a sampling-determinism bug (#91) — the shape of it (two targets, reproducible, identical content passing on main) looked like a code bug, and I filed it as circumstantial evidence there. It was wrong, and #91 now carries a correction. Defect 2 is precisely what made that misdiagnosis easy.

**This does not fix the soldr cache race itself** — it makes it fail loudly and in the right place. Retry/`--no-cache` handling stays in #92.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -66,9 +66,18 @@ jobs:
           # reading it, killing individual test binaries. The tool's own advice is to
           # retry, and to bypass the cache if that fails -- reruns alone do NOT help,
           # because the restored cache carries the same poisoned entry back in.
+          # Keep the raw JSON. Piping cargo straight into jq discards every
+          # `compiler-message` record, which is where the actual error text lives -- so a
+          # failing build printed only "could not compile (test ...)" with no reason, on
+          # every retry. Render the diagnostics on failure instead of dropping them.
           run_build() {
-            eval "$1 --target $TARGET --release --message-format=json" \
-              | jq -r 'select(.reason == "compiler-artifact" and .profile.test and .executable != null) | .executable' > bins.txt
+            local rc=0
+            eval "$1 --target $TARGET --release --message-format=json" > cargo.json || rc=$?
+            jq -r 'select(.reason == "compiler-artifact" and .profile.test and .executable != null) | .executable' cargo.json > bins.txt
+            if [ "$rc" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


zackees:
Merging with `--admin` while `cross` is red, with the justification stated rather than assumed.

**The failure is environmental and affects `main` equally.** Control experiment: I re-ran the previously-green `main` run (30672061072) with zero content changes — it now fails. Same commit, opposite result. Full evidence in #92.

So this PR cannot go green until the soldr/zccache regression clears, and blocking on that would leave the diagnosis fixes — which are what identified the problem in the first place — unlanded while every future `cross` failure stays equally opaque.

What this PR does *not* do is fix the underlying breakage; `cross` remains red repo-wide, tracked in #92 with the remaining ask (get zccache out of the build script's C compilation; `--no-cache` is insufficient because it bypasses soldr's cargo cache, not zccache).

PR #95, opened 2026-08-01
ci(cross): stop restoring soldr's build-cache, which is serving a broken C object

### LINKED ISSUE #92: cross: build job uploads a partial dist/ when compilation fails, hiding the cause

zackees:
Fixes #92 (the remaining half — the diagnosis half landed in #93).

## The failure

`cross` has been failing repo-wide since ~23:10 with:

```
undefined reference to `mi_prof_is_enabled'
undefined reference to `mi_prof_start_seeded'
undefined reference to `mi_prof_dump_writer'   (and every other mi_prof_*)
```

Ruled out as a code problem before touching CI:

- `build.rs:22` defines `MI_PPROF=1` **unconditionally**
- the vendored amalgamation contains every one of those symbols (`grep -c`: 3, 7, 5, 6)
- `rust-native` builds the same crate and passes

So the C object being linked is **incomplete**, not mis-configured — consistent with soldr#1969 ("an intermediate can be reclaimed while a slow link is still reading it").

## Confirmed environmental, by control experiment

I re-ran a **previously green `main` run** on byte-identical content. It now **fails**. Same commit, opposite result — no commit caused this.

Two things that do **not** fix it, both tried:

- **Reruns.** The restore brings the same poisoned entry back; three reruns failed identically.
- **`soldr --no-cache`.** The retry added in #93 fires and still fails, because `--no-cache` bypasses soldr's *cargo* cache while the build script's C compilation goes through **zccache**, seeded separately.

## The fix

`build-cache: false` on all three `setup-soldr` steps in this workflow. Per the action's docs that input controls "the Soldr/zccache compilation artifact cache" specifically — so the poisoned layer goes and **toolchain, registry, and mini caches stay on**. The cost is recompilation, not a cold toolchain across six targets.

Scoped to `cross` only; `rust-native` is unaffected and passing.

## This is also the experiment

Nothing else changes in this PR. **If cross goes green, that confirms the diagnosis**; if it stays red, the artifact cache was not the cause and #92 reopens with that ruled out.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -45,7 +45,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: zackees/setup-soldr@v0
         with:
-          cache: true
+          # cache:false is the action's documented kill switch -- it "disables every
+          # layer and exports environment variables to skip zccache wrapping entirely".
+          # build-cache:false alone was NOT enough: the C objects are still served by the
+          # in-run zccache daemon, so the missing-symbol object survived.
+          cache: false
           # setup-soldr now defaults to SOLDR_LINKER=fast, injecting mold/rust-lld in
           # place of the toolchain's linker. rust-lld does not pull the archive members
           # GNU ld does for our C static lib, so every test binary failed to link with
@@ -126,7 +130,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: zackees/setup-soldr@v0
         with:
-          cache: true
+          # cache:false is the action's documented kill switch -- it "disables every
+          # layer and exports environment variables to skip zccache wrapping entirely".
+          # build-cache:false alone was NOT enough: the C objects are still served by the
+          # in-run zccache daemon, so the missing-symbol object survived.
+          cache: false
           # setup-soldr now defaults to SOLDR_LINKER=fast, injecting mold/rust-lld in
           # place of the toolchain's linker. rust-lld does not pull the archive members
           # GNU ld does for our C static lib, so every test binary failed to link with
@@ -202,7 +210,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: zackees/setup-soldr@v0
         with:
-          cache: true
+          # cache:false is the action's documented kill switch -- it "disables every
+          # layer and exports environment variables to skip zccache wrapping entirely".
+          # build-cache:false alone was NOT enough: the C objects are still served by the
+          # in-run zccache daemon, so the missing-symbol object survived.
+          cache: false
           # setup-soldr now defaults to SOLDR_LINKER=fast, injecting mold/rust-lld in
           # place of the toolchain's linker. rust-lld does not pull the archive members
           # GNU ld does for our C static lib, so every test binary failed to link with
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -46,10 +46,18 @@ jobs:
       - uses: zackees/setup-soldr@v0
         with:
           cache: true
+          # build-cache restores soldr/zccache's compiled artifacts. A degraded
+          # entry there serves a C object with the mi_prof_* symbols missing, and
+          # every link then fails with `undefined reference to mi_prof_is_enabled`.
+          # Confirmed environmental: re-running a previously-GREEN main run on
+          # unchanged content now fails. Reruns cannot clear it, because the restore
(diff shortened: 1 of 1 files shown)


zackees:
Closing unmerged. Both hypotheses this PR tested were **disproved**, and shipping it would cost CI time for no fix.

Tested and ruled out, in order:

1. **`build-cache: false`** — still red. The GitHub Actions cache layer was not the cause.
2. **`linker: platform-default`** — still red. The driver changed from `clang` to `cc`, but `rust-lld` remained the actual linker, so soldr injects in more than one place.
3. **`cache: false`** (the action's documented "skip zccache wrapping entirely" kill switch) — still red.

So the `cross` failure is not soldr's caching and not its linker injection. What remains established:

- **not our code** — `build.rs` defines `MI_PPROF=1` unconditionally, the vendored amalgamation contains every missing symbol (`grep -c`: 3/7/5/6), and `main` is drift-free
- **environmental** — a previously green `main` run, re-run on byte-identical content, now fails
- **consistently target-split**, which random corruption cannot explain: `aarch64-unknown-linux-gnu` (zigbuild) and `x86_64-pc-windows-gnu` pass every time; the other four fail every time

The likely remaining suspect is the **soldr binary fetched at runtime** (`soldr=0.8.23`) rather than the action itself — the action's last release is v0.9.73 from 2026-07-15, well before the cutover. Pinning the soldr binary version, if the action supports it, is the next thing to try.

Recorded on #92, which stays open. The diagnosis improvements from #93 are already merged and are what made this bisectable at all.

Issue #88, opened 2026-07-31
CI: a C-only PR can merge with a stale vendored amalgamation and break main after merge


zackees:
A C-only PR can merge while leaving `rust/mimalloc-pprof/vendor/mimalloc-pprof-amalgamated.c` stale, and `main` goes red *after* the merge rather than on the PR.

This happened: #72 fixed a warning in `src/profile.c`, merged green, and broke `rust-native` on all three platforms. Fixed by #85.

## Why the existing setup allows it

Repo rule 2 forbids mixing C-core and `rust/` paths in one commit, so a C change **must** be followed by a separate Rust-only regeneration commit. That rule is correct -- it keeps upstream cherry-picks clean -- but nothing enforces the second half. The drift check lives in `xtask check`, which only runs in `rust-native`, and `rust-native`'s path filters mean a C-only PR may not trigger it at all.

So the guarantee is "the author remembers", which is exactly the kind of guarantee this repo has already learned not to rely on.

## Two failure modes it created

1. **Late detection.** The breakage surfaced on `main` post-merge, not on the PR.
2. **Misattribution.** The next PR to run `rust-native` inherits a red job it did not cause. I initially misread it as the known macOS `setup-soldr` EPIPE flake -- that EPIPE *was* present in the macOS log, but ubuntu and windows had failed for the entirely different DRIFT reason. One platform's log generalized to three.

## Ask

Make drift a **PR-blocking** check that runs whenever C sources change:

- add a cheap job (or a step in `c-unit`) that runs the drift check on any PR touching `src/**`, `include/**`, or `CMakeLists.txt`
- it must **fail**, not warn, and its message should name the fix command and the rule-2 reason for the separate commit
- it must not require the whole Rust toolchain if that makes it slow -- if `xtask check` is expensive, a cheaper equivalent (regenerate to a temp file, compare) is fine

## Related gap worth fixing at the same time

**A docs-only PR merges with zero checks.** #81 (README only) ran nothing, because every workflow's `paths:` filter excludes `README.md`. That is reasonable on its own, but it means a docs PR cannot surface an already-broken `main` either. Consider a minimal always-on job so every PR has at least one required check, rather than some PRs having none.

## Acceptance

- A C-only PR that leaves the vendored copy stale fails **on the PR**
- Demonstrated by pushing exactly that and linking the red run -- per the standing rule here that a gate never observed to fire proves nothing
- Recorded decision on whether every PR should have at least one required check

[...]

PR #96, opened 2026-08-01
rust: regenerate the vendored amalgamation after zero-tracking (#79)

zackees:
**Fixes a red `rust-native` on `v4`.**

#79 (zero-tracking) added `mi_option_purge_zeroes` to `include/mimalloc.h` and changed `src/arena.c` / `src/options.c`, but none of it reached the vendored copies. `xtask check` reports drift on all three:

```
DRIFT: vendor/mimalloc-pprof-amalgamated.c does not match a fresh amalgamation of src/static.c
DRIFT: vendor/mimalloc-pprof-amalgamated.h does not match a fresh amalgamation of the public headers
DRIFT: vendor/mimalloc.h does not match include/mimalloc.h
```

The diff is exactly #79's surface: the enum entry, its options-table row, and the guarded dirty-bit clear in the purge path. `xtask check` is now clean.

## Worth stating plainly

**This is the second time I have made this exact mistake** — the first broke `main` after #72 and was fixed in #85. Rule 2 (never mix C-core and `rust/` paths in one commit) means every C change needs a Rust-only follow-up commit, and forgetting it is invisible until a downstream job fails, often on a later PR that didn't cause it.

#88 tracks making CI catch this **on the PR** rather than after the merge. Until that lands this stays a manual step, and my having repeated it is the argument for prioritizing it.

### DIFF
--- a/rust/mimalloc-pprof/vendor/mimalloc-pprof-amalgamated.c
+++ b/rust/mimalloc-pprof/vendor/mimalloc-pprof-amalgamated.c
@@ -1,4 +1,4 @@
-/* GENERATED FILE -- DO NOT EDIT. Produced by rust/xtask from commit be5e067f of src/static.c. Regenerate with: cargo run -p xtask -- amalgamate-c */
+/* GENERATED FILE -- DO NOT EDIT. Produced by rust/xtask from commit fba9f8cc of src/static.c. Regenerate with: cargo run -p xtask -- amalgamate-c */
 
 /* ---- begin inlined: src/static.c ---- */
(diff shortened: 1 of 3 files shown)


zackees:
Merging with `--admin`. **`rust-native` — the job this PR fixes — is green on all four platforms.** The only red checks are `cross`, which is failing repo-wide on every branch including `main` for an unrelated environmental reason (#92: three hypotheses tested and disproved in #95; a previously green `main` run re-run on identical content also fails).

Blocking on `cross` would leave `v4`'s `rust-native` red for a defect this PR actually fixes.

Issue #86, opened 2026-07-31
ASan smoke test: we have an ASan job that has never run on any of our branches


zackees:
Requested as: "a smoke test that uses asan to root out bugs."

**The important finding, checked before filing: we already have an ASan job, and it has never run.**

## Current state

`.github/workflows/test.yaml:129-131` — inherited from upstream — does exactly the right thing:

```sh
CC=clang CXX=clang++ cmake . -B out/debug-asan -DCMAKE_BUILD_TYPE=Debug -DMI_DEBUG_FULL=ON -DMI_TRACK_ASAN=ON
cmake --build out/debug-asan --parallel 8 --config Debug
ctest --test-dir out/debug-asan --verbose --timeout 240 -C Debug
```

But its trigger is:

```yaml
on:
  workflow_dispatch:
  push:
    branches: ['dev*']
    tags: ['v*']
```

**Our branches are `main`, `v2`, and soon `v4`. None match `dev*`.** There is no `pull_request` trigger at all. So this job has never run on a single commit of this fork — it is inherited configuration that looks like coverage and provides none.

The CMake support is real and present (`MI_TRACK_ASAN`, plus `MI_DEBUG_UBSAN` and `MI_DEBUG_TSAN` at `CMakeLists.txt:22-25`). Only the wiring is missing.

## Why this matters more for us than for upstream

Upstream's ASan job covers upstream's code. **It does not cover ours**, and our additions are exactly the kind that ASan is good at:

- `src/profile.c` maintains an intern table and a linked list of sample records with manual lifetime management
- profiler memory comes from `_mi_os_alloc` (rule 4), so it sits **outside** the allocator's own redzones and debug fills — mimalloc's internal checks do not see it
- `src/memory-events.c` has `mi_unwrapped_*` with a header-before-pointer layout, which is a classic underflow shape
- `test-profile-race.c` deliberately drives concurrent dump/free paths

None of that existed when the inherited job was written.

## Ask

1. **Wire a real ASan job into `c-unit`** (or its own workflow) triggering on `pull_request` and `push` to our branches. Ubuntu + clang is the cheap baseline; macOS ASan conflicts with `MI_OVERRIDE=ON` per `CMakeLists.txt:292`, so scope it deliberately rather than discovering that in CI.
2. **Cover the profiler paths specifically** — `test-profile`, `test-profile-race`, `test-memory-events`, `test-degenerate` — not just the upstream API tests.
[...]

Issue #80, opened 2026-07-31
Bump the upstream pin from 579f8c0e to the dev3 tip


zackees:
Acting on the C.1 decision recorded in #66.

## Why now

Measured in #66 rather than estimated — the overlay was applied to the tip and run:

```
gap                        92 commits (579f8c0e -> d077de56)
12 of 12 hook patches      git apply --check CLEAN, zero conflicts
build (MinGW, MI_PPROF=ON) 0 errors
ctest                      13/13, including test-degenerate (the leak detector)
```

The cost is near zero **right now**, and that is the argument for moving: 92 commits of drift that cost nothing to absorb today will cost more at 292. Staying pinned only banks a larger bill.

## Scope

1. Move the pin to `upstream/dev3` tip (re-check the tip SHA at the time of work; `d077de56` was current when #66 was measured).
2. Re-apply the fork hooks and confirm each still lands where intended — *applies cleanly* is not the same as *still correct*. `src/alloc.c` (91 lines) and `src/free.c` (88) churned most, and that is exactly where the memory-events hooks live: re-read those hunks in context rather than trusting the patch.
3. Regenerate `src/static.c` if the file list changed, and the Rust vendored amalgamation via `rust/xtask` — in a **separate Rust-only commit** per repo rule 2.
4. Full `c-unit` matrix green: ubuntu/windows-MSVC/windows-MinGW/macos, `MI_PPROF` ON and OFF, plus `memory-gate` and `isa-baseline`.
5. Re-run the memory-gate baselines afterward if the allocator's legitimate peak moved, and say so explicitly rather than silently re-baselining.

## Watch for

- `prim/windows/prim.c` had **zero** upstream churn, so the MinGW TLS fix should carry unchanged. If it suddenly conflicts, upstream changed the init path and that needs reading before resolving.
- Upstream still has not fixed the MinGW thread-exit leak (#56 / `docs/upstreaming.md`), so do not assume the bump makes our patch redundant. Verify with `test-degenerate`, not by inspection.
- The `pr/*` upstreaming branches are cut from upstream history; a pin bump may make some of them rebaseable or redundant. Check before re-cutting.

## Acceptance

- Pin moved, full matrix green
- Each hook hunk confirmed in its new context, not just applied
- Any baseline change called out with the reason

Refs #66, #61

zackees:
**This is the first v4 change** — see #61 (rev 4). Everything else in the v4 scope rebases onto it, so it goes in before #78.

[...]

Issue #74, opened 2026-07-31
Add pyright + ruff to the Python CI scripts and make them type-safe


zackees:
The repo's Python is small but load-bearing: it is the CI gating layer. `ci/memory_gate.py` decides whether a PR is a memory regression and `ci/check_isa_baseline.py` decides whether a build is portable. **A bug in these scripts silently disables a gate** — exactly the failure mode we already hit twice this week:

- the arm64 ISA scanner matched nothing and reported "clean" (caught only by a positive control)
- the memory gate's missing-baseline path returned before printing its diagnostics

Neither was a C bug. Both were Python.

## Current surface

| File | Lines | Role |
|---|---|---|
| `ci/dev_linux.py` | 186 | local dev loop (`c-test` / `rust-test` / `bench`) |
| `ci/memory_gate.py` | 229 | memory regression gate |
| `ci/check_isa_baseline.py` | 216 | ISA portability gate |

~630 lines, no linting, no type checking, no CI job.

## Proposed

1. **`ruff`** — lint + format. Fast, single binary, no plugin zoo. Enable at least `E,F,W,I,B,UP,SIM,RUF`.
2. **`pyright`** in `strict` mode — real type checking. These scripts parse JSON and index dicts by string key; `strict` is what catches a typo'd key or a `None` that reaches arithmetic.
3. **Type-safety work**, which is the substantive part rather than just adding config:
   - JSON payloads are currently untyped `dict[str, Any]`. Give the result/baseline schema a `TypedDict` so `result["peak_mb"]` is checked, not hoped for.
   - `check()` returns bare `int` exit codes; an `IntEnum` or `Literal[0,1,2]` makes the contract explicit and stops a `return 2` meaning "regression" by accident.
   - `subprocess` and file IO paths need explicit `str`/`Path` discipline.
4. **CI job** `python-lint` running both, on the same triggers as `c-unit`, as a **hard failure**. Non-negotiable given these files gate other jobs.
5. **`pyproject.toml`** to hold the config (there is none today), pinned tool versions so CI is reproducible.

## Acceptance

- `ruff check` and `ruff format --check` clean
- `pyright --strict` clean (no blanket `# type: ignore`; suppressions need a reason)
- CI job green and required
- No behavior change to the gates — verified by re-running each gate's positive control in both directions afterward, since a refactor that quietly breaks a gate is the specific risk here

## Ordering

[...]

Issue #65, opened 2026-07-31
Phase B2: decide the MIMALLOC_PROF* env namespace


zackees:
Sub-issue of #61 (Phase B).

`MIMALLOC_*` environment variables are **process-global**, and **NVIDIA's display driver embeds mimalloc 3.1.6** (found via the Arma 3 fork, which renamed its vars to `ARMA3_MIMALLOC_*` for exactly this reason). So our `MIMALLOC_PROF*` variables are visible to NVIDIA's instance and vice versa.

Renaming after users depend on the names is a breaking change, so this is decided **before** 0.9.x sees wider adoption or not at all.

### Options
1. Keep `MIMALLOC_PROF*` — collision risk accepted; NVIDIA's instance ignores unknown vars, so the practical risk is our profiler reacting to a var someone set for a different mimalloc.
2. Prefix to `MIPPROF_*` or similar — clean, but breaks the 0.9.0 API we just published.
3. Accept both, preferring the prefixed form, and document.

### Acceptance
A decision recorded here, then a mechanical rename if needed. One decision, not a research task.

Refs #61

PR #99, opened 2026-08-01
release: mimalloc-pprof 0.9.1

### LINKED ISSUE #82: Release 0.9.1: ratchet the version so the #61 hardening ships
Everything landed under #61 is on `main` but **not in any published crate**. `rust/mimalloc-pprof/Cargo.toml` is still `0.9.0`, which is the release from before this work. Users on crates.io have none of it.

## What is on main but unreleased

| Change | Issue / PR |
|---|---|
| Memory + counter regression gate, with per-platform baselines and a working positive control | #63 / #70 |
| ISA baseline check + portable-build artifact; fixed a scanner that reported "clean" while matching nothing on arm64 | #64 / #71 |
| Shared-library CI coverage; fixed 4 test targets that hardcoded `mimalloc-static` and broke shared-only builds | #62 / #68 |
| `MIMALLOC_PROF*` namespace decision + `MI_PROF_CONFIG_OVERRIDE` tested across every env-backed field | #65 / #73 |
| `-Wmaybe-uninitialized` fix in `src/profile.c` | #72 |
| `ruff` + `pyright --strict` over the CI gate scripts | #74 / #76, #77 |

Note the CMake fix (#68) and the profile.c warning fix (#72) are **user-visible**: anyone building shared-only was broken.

## Proposed: 0.9.1

Patch rather than minor -- no API was added or changed. `mi_option_purge_zeroes` (#67 / #79) would be new public API, so **if that merges first, this becomes 0.9.2/0.10.0** and the decision moves with it.

## Checklist

- [ ] Bump `rust/mimalloc-pprof/Cargo.toml` (and the sys crate if versioned separately)
- [ ] Decide whether `MI_MALLOC_VERSION` (currently `30403`) needs to move -- it tracks the *upstream* engine version, so probably not, but state the decision rather than leaving it implicit. Related: #80 may change it.
- [ ] Regenerate the vendored Rust amalgamation, in a Rust-only commit (repo rule 2)
- [ ] Dry-run the publish (via `soldr`, per repo convention) and confirm it is clean, including the metadata that bit us before: repository / homepage / documentation / crate-level README
- [ ] Full `c-unit` matrix + `rust-native` green
- [ ] Release notes leading with the shared-build fix, since that is the one that unbreaks someone
- [ ] Verify the published crate actually installs from crates.io afterward, rather than trusting the publish step's exit code

## Ordering

After #79 resolves (merge or drop), so the version number reflects whether new public API shipped. Not blocked on #80 (pin bump) -- that should be its own release.

Refs #61

zackees:
Closes #82.

Everything from the #61 hardening pass is merged but **unpublished** — crates.io still serves 0.9.0, from before that work. This bumps the crate so it actually reaches users.

## Version decisions, stated rather than implied

- **0.9.1, not 0.10.0** — no API was added or changed. `mi_option_purge_zeroes` (#67) *is* new public API, but it landed on `v4`, not here.
- **`MI_MALLOC_VERSION` stays at `30403`.** It tracks the *upstream engine* version, and the pin is unchanged (#80 still open), so moving it would misreport which mimalloc actually ships.

## Contents, leading with the fix that unbreaks someone

- **Shared-library builds were broken.** Four test targets hardcoded `mimalloc-static`, so `-DMI_BUILD_SHARED=ON -DMI_BUILD_STATIC=OFF` failed to link with `cannot find -lmimalloc-static` (#62, #68).
- `-Wmaybe-uninitialized` on `fmt_buf` in `src/profile.c` (#72).
- `MI_PROF_CONFIG_OVERRIDE` now tested against a hostile ambient environment for **every** env-backed field (#65, #73).
- New CI gates, each with a positive control: memory/counter regression gate (#63), ISA baseline scanner (#64), **AddressSanitizer** (#86), amalgamation drift guard (#88), `ruff` + `pyright --strict` (#74).

## ASan validated first, per the stated goal

Publishing was gated on ASan actually working, not merely existing. #98 landed it and the run confirms all three layers:

```
-- Compile with address sanitizer support (MI_TRACK_ASAN=ON)
ASan confirmed linked into build-asan/mimalloc-test-api
100% tests passed, 0 tests failed out of 13
==8132==ERROR: AddressSanitizer: use-after-poison ... in main
positive control fired: ASan caught the use-after-free (rc=1)
```

**13/13 under ASan with no findings**, and the control proves mimalloc is poisoning freed blocks — i.e. ASan sees our allocations, not just that it is linked.

## Release procedure

Merge this, then push the `v0.9.1` tag, which is what `auto-release.yml` treats as a real release. A `workflow_dispatch` dry run goes first.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -35,7 +35,7 @@ checksum = "cf8baf1c55e62ffcace7a9f06f4bd9cd3f0c4beb022d3b367256b91b87513d98"
 
 [[package]]
 name = "mimalloc-pprof"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "cc",
  "regex",
--- a/rust/mimalloc-pprof/CHANGELOG.md
+++ b/rust/mimalloc-pprof/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## 0.9.1
+
+Hardening release. No API changes -- everything here is a fix or a CI gate.
+
+- **Fix: shared-library builds were broken.** Four test targets hardcoded
+  `mimalloc-static`, so a `-DMI_BUILD_SHARED=ON -DMI_BUILD_STATIC=OFF` build failed to
+  link with `cannot find -lmimalloc-static` (#62, #68). This is the one that unbreaks
+  someone.
+- Fix: `-Wmaybe-uninitialized` on `fmt_buf` in `src/profile.c` (#72).
+- `MI_PROF_CONFIG_OVERRIDE` is now tested against a hostile ambient environment for
+  *every* env-backed field, not just `sample_interval` (#65, #73). The documented way to
+  be immune to process-global `MIMALLOC_*` is now covered by tests rather than by
+  assertion.
+- New CI gates, each with a positive control proving it can actually fail: peak-memory
+  and thread-counter regression gate with per-platform baselines (#63), CPU-baseline
+  instruction scanner for portable builds (#64), AddressSanitizer with an
+  allocator-mediated use-after-free control (#86), vendored-amalgamation drift guard
+  (#88), and `ruff` + `pyright --strict` over the CI scripts (#74).
+
+Four of those gates were found to be silently checking nothing when first written; the
+controls are why that was noticed. See the README's "CI gates" section.
+
 ## 0.8.0
 
 First real release of `mimalloc-pprof`.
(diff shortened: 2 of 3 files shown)


zackees:
Merging with `--admin`. Every check that validates *this* change passes — `rust-native` on all four platforms including `cargo publish --dry-run` for 0.9.1, plus the amalgamation drift guard.

The red checks are all `cross`, failing repo-wide on every branch including `main` for an unrelated environmental reason (#92 — three hypotheses tested and disproved in #95; a previously green `main` run re-run on identical content also fails).

Issue #69, opened 2026-07-31
MSVC shared build: MIMALLOC_MEMORY_EVENTS env activation fails (lazy resolution races DLL init)


zackees:
Found by the new shared-library CI job (#62 / PR #68). **Windows MSVC shared build only** — ubuntu shared, win-gnu shared, and all static builds pass.

```
12/13 Test #12: test-memory-events-env-enabled ...***Failed  0.01 sec
92% tests passed, 1 tests failed out of 13
```

Note `test-memory-events` (test 11) **passes** on the same DLL, so the API works and the DLL loads correctly. Only the environment-variable activation path fails, and it fails in 0.01 s.

## Hypothesis

`src/memory-events.c` resolves the environment **lazily, on first allocation**:

```c
static void memevt_resolve_env(void) {
  if (_mi_atomic_once_enter(&memevt_once)) {
    const bool enabled = mi_option_is_enabled(mi_option_memory_events);
    mi_atomic_store_release(&memevt_state, ...);
```

and the test's own comment states the assumption explicitly:

> Nothing must have touched the allocator yet in this fresh process; the very first allocation hook below is what triggers the lazy env resolution.

**In an MSVC DLL build that assumption is false.** Allocations occur during `DLL_PROCESS_ATTACH` and CRT initialization, before `main`. If `mi_option_is_enabled` at that point has not yet read the environment, the `once` caches **disabled permanently**, and by the time `main` checks, tracking can never turn on.

That is the same bug class as `mi_heap_new`/`mi_subproc_new` failing to bootstrap (#43/#44) and upstream #1341 — **an initialization-ordering assumption that holds for static links and breaks for dynamic ones**. It is also why the static suite never caught it: in a static build the first allocation genuinely does happen in `main`.

## Not yet verified

I do not have MSVC locally and could not reproduce it outside CI, so the above is a hypothesis consistent with the evidence, not a confirmed root cause. Do not fix it on my say-so — confirm first.

## Suggested investigation

1. Add stderr output to the failing path so CI shows *which* of the five return codes it hits — that alone distinguishes "env not visible" from "tracking not enabled" from "callback did not fire".
2. Check whether `mi_option_is_enabled` is safe to call at `DLL_PROCESS_ATTACH` on MSVC, and whether options have read the environment by then.
3. If confirmed: either resolve the env at process init rather than lazily, or make the `once` re-resolvable until the first explicit API call.

## Status of the CI job

[...]

PR #112, opened 2026-08-01
fix: MSVC shared build MIMALLOC_MEMORY_EVENTS activation (#69)

zackees:
Working #69. **First commit deliberately adds the failing job to capture evidence** — the root cause is explicitly unverified and I have no MSVC locally, so this diagnoses before fixing.

The test already distinguishes five failure modes by stderr (returns 1–5: env not set / alloc failed / tracking not enabled / second alloc failed / callback did not fire), and `--output-on-failure` is already set. What was missing was simply a run.

**Expected red on the first commit.**

### DIFF
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -97,6 +97,19 @@ jobs:
       - run: cmake -B build -DMI_PPROF=ON -DMI_BUILD_SHARED=ON -DMI_BUILD_STATIC=OFF -DMI_BUILD_OBJECT=OFF
       - run: cmake --build build --config Release
       - run: ctest --test-dir build -C Release --output-on-failure
+      # DIAGNOSTIC (#69): ctest reported the env-activation test as Failed with no output
+      # at all, despite --output-on-failure, so the five distinct stderr messages the test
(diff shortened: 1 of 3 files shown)


zackees:
Closing unmerged. The attempted fix (not latching the env read during `_mi_preloading()`) **did not resolve it** — MSVC shared still exits 3 — so nothing speculative should land on an allocation hot path.

Full diagnosis recorded on #69, including the part where my own probe misled me: a Git-bash-launched run exited 1 (`getenv` NULL) and I briefly reported that as refuting the hypothesis. Running via `cmd.exe` gave exit 3 — `tracking not enabled after lazy env read` — which *supports* it. The probe was broken, not the hypothesis.

What this PR did establish, and what the next attempt inherits:

- **The failure mode is now known**: env visible, CRT up, `mi_memory_tracking_is_enabled()` still false after the first allocation.
- `ctest --output-on-failure` produces **no output at all** for this test on MSVC — running the binary directly is required to see anything.
- Ruled out: `MI_NO_GETENV`, the `_mi_prim_getenv` errno translation, and the preloading-latch theory as a sufficient explanation.

The MSVC-shared job stays out of the matrix — it is red for a real reason, and adding it red is how a gate becomes decoration.

PR #204, opened 2026-08-14
feat(benchmarks): sparse thread-scaling sweep with dark README panels

### LINKED ISSUE #203: feat(benchmarks): sparse scaling sweep - 1/4/16/64 threads x 4 seeded random allocation patterns
Parent phase: #183
Implementation guide: #177
Design authority: #176 (this issue records an owner-directed amendment)
Supersedes the measurement protocol of: #186
Branch: `feat/benchmark-sparse-scaling-sweep`
PR count: exactly one

## Context

The current daily core-throughput run burns ~32 minutes of measurement (30 cells x 4 allocators x 15 blocks x ~1 s calibrated blocks; run 31586930232 spent 31m54s in `run benchmark suite` out of a ~40-minute job). The dense Phase 6C scaling protocol in #186 (6 topology-resolved thread points, >=15 blocks per point, <=60-minute budget) would add another long-burn workflow.

Owner decision (2026-08-13): **sacrifice statistical rigor for thread-scaling coverage.** Replace the dense 6C protocol with a sparse sweep at fixed numeric thread counts **1, 4, 16, 64**, crossed with **4 allocation patterns**, each driven by a repeatable seeded random operation sequence.

Budget math for the new mode: 4 patterns x 4 thread points = 16 cells, x 4 allocators x 3 blocks x ~0.75 s target = ~144 s of pure measurement; with calibration probes, child spawn, warmup, and teardown the job should land **under 15 minutes end to end** — versus 32+ minutes for the current suite and versus the ~60 minutes #186 budgeted.

### 1. Metric `throughput-scaling-sparse-v1`

New metric version, published through the same overlay path Phases 6A/6B established (separate workflow writes into `latest.json`/`history.jsonl` on `benchmark-stats`, replacing the pending `benchmark-scaling.png` panel).

Thread points are **fixed numeric 1, 4, 16, 64** — no topology resolution, no deduplication logic, no affinity pinning requirement. Record runner CPU topology (allowed cpuset, physical/logical counts) as metadata only, and label each point with its oversubscription factor relative to the runner (on the 4-vCPU `ubuntu-24.04` runner: 1 and 4 are at-or-below capacity; 16 is 4x and 64 is 16x oversubscribed). The chart must show these labels; oversubscribed points are honest data about contention behavior, not fake core scaling.

### 2. Four allocation patterns as seeded random operation streams

Unlike the fixed-recipe cards in `rust/benchmark-suite/src/scenarios.rs`, each pattern is a **randomized operation sequence**: a deterministic PRNG stream (the suite's existing splitmix64) draws each operation (malloc / free-oldest / free-random / realloc), each size from the pattern's distribution, and each slot from a bounded live-set, so the workload has irregular interleaving instead of lockstep alloc-touch-free loops.

| Pattern id | Sizes | Character |
|---|---|---|
| `sparse-tiny-hot` | 16-64 B | small-object hot path, high alloc/free rate, small live set |
| `sparse-mixed-general` | 8 B-4 KiB log-uniform | general-purpose mix incl. ~10% realloc, medium live set |
| `sparse-large-buffers` | 64 KiB-4 MiB | large allocations with page touching, small live set |
| `sparse-cross-thread` | 16-512 B | producer/consumer handoff: blocks freed by a randomly-chosen *other* worker (remote-free pressure) |

Seeding contract (repeatability requirement):

- one master `run_seed` recorded in the raw artifact (derived or explicit, same as `benchmark-run --run-seed` today);
- per-stream seed = splitmix64 chain over (run_seed, pattern id, thread point, block, worker id);
- **the same worker id replays the identical operation stream across all 4 allocators within a paired block** — preserving the paired-comparison property even in coverage mode;
- replaying with the recorded seed reproduces the exact operation sequence (existing `reproduction_command` mechanism extends to these children).

### 3. Reduced statistical protocol (explicit downgrade)

- **3 blocks** per (pattern, thread point, allocator) cell — not 15;
- per-block target ~0.5-1 s via the existing upstream-child calibration, calibrated once per (pattern, thread point) and frozen across allocators;
- fixed per-worker operation count across allocators at a given thread point (total ops scale with worker count, matching #186's work-normalization rationale);
- report median and min/max across the 3 blocks; **no bootstrap CIs, no noise gating** — the validator only enforces completeness (all 16 cells x 4 allocators x 3 blocks present) and determinism, not statistical significance;
- the published panel and HTML carry a visible **"coverage mode — reduced statistical rigor (3 blocks)"** label so nobody mistakes these for the headline-grade numbers.

### 4. Workflow and burn-time recovery

- New `benchmark-scaling.yml` cloned from the Phase 6B `benchmark-latency.yml` shape (workflow_dispatch + weekly cron, same concurrency group `benchmark-stats-production`, same sealed publication/audit path, least-privilege permissions), **hard timeout 20 minutes**;
- structural policy checker `ci/check_benchmark_scaling_workflow.py` + tests, matching the 6A/6B pattern;
- **cadence change to recover the daily burn:** the sparse sweep runs daily alongside a *weekly* full-rigor core-throughput run — the daily `benchmark-stats.yml` cron moves to weekly, and the sparse sweep becomes the daily freshness signal. History continuity is unaffected: fewer rows per comparison key, same key.

### 5. Chart

Replace the pending `benchmark-scaling.png`: one facet per pattern, x-axis = thread count (1/4/16/64, log2 spacing), y-axis = aggregate throughput, one line per allocator, oversubscription region visually marked, coverage-mode label, runner topology in the caption.

## Acceptance criteria

- [ ] RED: failing tests exist before implementation for (a) stream determinism — same (seed, pattern, thread point, block, worker) reproduces the identical operation sequence and differing worker/allocator inputs do not, (b) cross-allocator stream identity within a paired block, (c) validator rejection of an incomplete 16-cell x 4-allocator x 3-block matrix, (d) schema validation of the new scaling section.
- [ ] GREEN: implementation turns all of the above green; full local gates pass (`soldr cargo test -p benchmark-suite --locked`, workflow policy checker selftests, `ruff`/`pyright`, rendering tests).
- [ ] All four patterns implement the seeded random operation stream with recorded seeds and a working reproduction command.
- [ ] Thread points are exactly 1, 4, 16, 64 with oversubscription factors recorded and rendered.
- [ ] One full `benchmark-scaling.yml` run from the default branch completes in **under 20 minutes** and publishes: `benchmark-scaling.png` replaced, scaling section present in `latest.json`, branch/Pages digests equal. Post the run URL, runtime breakdown, and digests here.
- [ ] Coverage-mode (reduced rigor) labeling is visible on the panel, in `index.html`, and in the metric metadata.
- [ ] Daily/weekly cadence swap for `benchmark-stats.yml` is merged in the same PR and reflected in its policy checker.
- [ ] #186 is closed as superseded (or edited to point here) once the panel is live.

## Decisions

- **Supersede #186 rather than amend it.** The owner's directive (sparse 1/4/16/64, rigor traded for coverage) contradicts #186's core contract (topology-resolved points, >=15 blocks). A clean issue avoids a Frankenstein spec; #186 stays open until this panel is live, then closes as superseded.
- **3 blocks, median-only.** The minimum that still allows a paired comparison and catches a single wild outlier; going to 1 block would make any noise spike indistinguishable from signal.
- **Keep pairing and fixed per-worker work.** These cost nothing in wall-clock and preserve the one property that makes cross-allocator lines comparable; the rigor being sacrificed is sample count and CI machinery, not fairness.
- **New pattern streams instead of reusing scenario cards.** The user asked for a repeatable random operation sequence; the existing cards are fixed recipes. Four new streams cover the same behavioral axes (tiny hot path, mixed general, large, cross-thread) with irregular interleaving.
- **Fixed numeric thread counts, no affinity.** 1/4/16/64 as literal worker counts keeps the protocol runner-independent; topology is recorded but not enforced. Oversubscribed points are labeled, not dropped.
- **Daily sparse + weekly full cadence.** This is the actual burn-time fix: daily cost drops from ~40 min to ~15 min while the rigorous suite still produces weekly headline data. Reversible by editing two cron lines if the daily full run is missed.
- **New metric version `throughput-scaling-sparse-v1`.** A different protocol must not share a version string with #186's never-shipped `throughput-scaling-v1`; comparison-key separation prevents chart lines from mixing protocols.

## Open questions

- Whether 64 threads on a 4-vCPU runner produces stable enough numbers even for coverage purposes, or whether the top point should carry an additional "informational only" marker after the first live run's min/max spread is observed.
- Whether the weekly full-rigor run should later absorb these four patterns at >=15 blocks (a possible Phase 6C-bis), or whether sparse coverage permanently replaces dense scaling.

## Related issues

- #186 — dense Phase 6C scaling protocol; this issue supersedes its measurement design.
- #183 — phase 6 coordinator; the scaling slot in its ordered chain is satisfied by this issue instead of #186.
- #176 / #177 — design authority and implementation guide; the cadence and rigor amendment recorded here is owner-directed.

zackees:
Implements #203 using the reduced matrix from that issue's [alternative-design follow-up](https://github.com/zackees/mimalloc-pprof/issues/203#issuecomment-5287429050) (**Alt B**): 4 allocation patterns × **3** thread points (1/4/16) × 4 allocators × 3 blocks = 144 samples.

The 64-thread point from the original proposal is dropped. It was the issue's own flagged-unstable cell; removing it cuts measurement 25% and eliminates the open question about its stability, while keeping all four access-pattern graphs the owner asked for.

## What lands

**Metric `throughput-scaling-sparse-v1`** (`rust/benchmark-suite/src/scaling.rs`). Each pattern is a seeded random operation stream, not a fixed recipe. A `WorkerPlanner` owns all randomness and live-set bookkeeping and never observes allocator behavior, so the executor and the derived oracle consume one state machine and cannot drift. The seed chain deliberately excludes the allocator, which makes cross-allocator stream identity inside a paired block a structural property rather than a runtime check.

| Pattern | Sizes | Character |
|---|---|---|
| `sparse-tiny-hot` | 16–64 B | small-object fast path, small live set |
| `sparse-mixed-general` | 8 B–4 KiB log-uniform | general mix including realloc |
| `sparse-large-buffers` | 64 KiB–4 MiB | large buffers, one byte per page touched |
| `sparse-cross-thread` | 16–512 B | producer/consumer handoff, remote free |

`sparse-cross-thread` hands blocks to a randomly chosen *other* worker through a bounded mailbox. A producer facing a full mailbox frees the block itself instead of blocking, so 16× oversubscription cannot livelock; the fallback count is published. Its checksum is a commutative sum of allocation-side touches, so the oracle stays exact even though completion interleaving is not fixed.

**Coverage mode is explicit and enforced**: 3 blocks, median with min/max, no bootstrap intervals, no noise gating. The rigor label is checked by both validators, so a report cannot be published claiming headline quality.

**Four dark SVG panels**, one per pattern, embedded in the README and on the dashboard.

## Why SVG instead of the existing PNG panel

The existing renderer is a hand-rolled PNG encoder with a rectangles-and-lines canvas and **no text rendering at all** — the live panels today carry no axis labels, tick values, or legend. A scaling chart needs a labelled log2 thread axis, a legend, and a visible coverage-mode caption, so `benchmark-scaling.png` is replaced by four self-contained SVGs. They are validated as inert (no script, no external fetch, no event handlers) because they are served from a raw branch URL, and the dark background is baked in so a panel does not inherit a host theme.

## Two pre-existing gaps fixed here

- **None of the `ci/check_benchmark_*.py` policy checkers ran in any workflow.** A regression in the policy guarding a job with `contents: write` and `pages: write` could not fail anything. They now run in `python-lint`, along with the `ci/tests` suite.
- **The 6A/6B checkers' `--selftest` is a rubber stamp** — it re-validates the real workflow and prints PASS. The new scaling checker's `--selftest` mutates a copy of the on-disk workflow once per rule and requires all **17** mutations to be rejected.

## Deliberately not included

The daily/weekly cadence swap for `benchmark-stats.yml`. This workflow lands weekly-first so one live run can prove the runtime budget before the daily signal depends on it; the swap is two cron lines in a follow-up.

## Verification

- Full Rust suite green, including 14 scaling contract tests. The key one re-derives every count and checksum from the seed chain and compares against real execution for all 4 patterns × 3 thread points.
- `pytest ci/tests`: 59 passed. `ruff check`, `ruff format --check`, `pyright`: clean.
- All four workflow policy selftests pass.
- Full local pipeline rehearsal: fixture → `benchmark-scaling-validate` → overlay onto `latest.json` → `render` → `validate-site`, with the sealed manifest digest verified.

A `/clud-review` pass found one HIGH and three MEDIUMs, all fixed in 82f127b5 — most importantly a barrier deadlock where a worker returning early stranded its peers, which would have surfaced in CI only as an opaque child timeout. A regression test now fails an allocator mid-run.

Closes #203.


zackees:
## CI status: three red workflows, all pre-existing

`python-lint` (which now gates the four benchmark policy checkers and the whole `ci/tests` suite), `c-unit`, and `amalgamation-drift` are green. Three workflows are red, and none of them are caused by this branch.

I did not want to assert that from "it looks unrelated", so I ran a control: a branch cut from `main` at `7f3bfef9` with a **single added comment line** in `rust/benchmark-suite/src/lib.rs` and no other change (PR #205, now closed). It reproduces all three failures.

| Workflow | This PR (run 31758615680 / 31758615…) | Control from `main` (run 31759288988) | Same cause |
|---|---|---|---|
| `rust-native` | test-win-gnu, windows, macos fail in `benchmark-suite/tests/timed_allocation_audit.rs` | same three jobs, same file | yes |
| `cross` | all 6 targets fail `planted_serialized_control_is_rejected` (`bench-harness/tests/planted_control.rs:38`) | same | yes |
| `benchmark-sentinel` | windows job fails before compiling anything: `soldr broker is unreachable` / `daemon route unavailable ... ErrorBackendSpawnFailed` | same | yes |

Detail on each:

**`rust-native`** — `timed_allocation_audit.rs` asserts zero harness-side allocations inside the measured interval, using a process-global counter and a global-allocator wrapper. On `main` this has been failing on macos and windows since #202 landed (`latency_sampling_buffers_do_not_allocate_after_warmup`, added by the 6B latency work); it has now spread to `test-win-gnu` on the current runner image. Which of the file's two tests trips first varies with scheduling — this PR's run reported `ordinary_measured_region_performs_zero_harness_allocations` while the control reported `latency_sampling_buffers_do_not_allocate_after_warmup` — but the control proves the file is red without any of this branch's code. Filed as #206.

**`cross`** — `planted_serialized_control_is_rejected` lives in `bench-harness`, a crate this PR does not touch, and fails identically on `main`'s own run 31553123529 from 2026-08-12.

**`benchmark-sentinel`** — the Windows job never reaches this branch's code; the soldr broker cannot spawn its daemon, so `unicode-ident`, `proc-macro2`, `quote`, and `serde_core` all fail to compile. This is the same soldr v0.9.0 failure I hit on my local Windows machine while developing (it is why the local Rust verification in this PR ran against the pinned rustup toolchain directly rather than through soldr). Environmental.

### Verification that did run for this branch

- Full `benchmark-suite` test suite green locally on Windows against the pinned 1.94.1 toolchain, including 14 new scaling contract tests. The load-bearing one re-derives every count and checksum from the seed chain and compares against real threaded execution for all 4 patterns x 3 thread points, so executor/oracle drift cannot pass.
- `rust-native` **ubuntu** — the platform this Linux-only metric actually runs on — is **green** on this PR.
- `pytest ci/tests`: 59 passed. `ruff check`, `ruff format --check`, `pyright`: clean. All four workflow policy selftests pass, including the new one's 17 negative controls.
- Full local pipeline rehearsal end to end: fixture -> `benchmark-scaling-validate` -> overlay onto `latest.json` -> `render` -> `validate-site`, with the sealed manifest digest verified.

PR #205, opened 2026-08-14
test: control run for rust-native win-gnu baseline

zackees:
Temporary control. Verifies whether rust-native test-win-gnu is red on unmodified main with the current runner image, independent of #204. Will be closed once the comparison is recorded.


zackees:
Control complete. Unmodified `main` plus a single comment line reproduces all three failures seen on #204 (rust-native test-win-gnu/windows/macos, cross, benchmark-sentinel), confirming they are pre-existing and not caused by that PR. Evidence recorded on #204; closing.

Issue #215, opened 2026-08-14
infra(benchmarks): run the thread-scaling sweep on real cores (8-16 vCPU)


zackees:
Split out of #211 WP4, which asked for three things: a dense thread sweep, a RSS-vs-thread-count curve, and *moving scaling onto a larger runner so the density is meaningful*. The first two landed with #211. This one is blocked on infrastructure, not code.

## Why it is blocked

| Check | Result |
|---|---|
| `gh api repos/zackees/mimalloc-pprof --jq .owner.type` | `User` — GitHub-hosted larger runners require an **organization** on Team/Enterprise. They cannot be provisioned for a personal account. |
| `gh api repos/zackees/mimalloc-pprof/actions/runners --jq .total_count` | `0` — no self-hosted runner is registered, so the `self-hosted-informational` / `stable-host` classes that `ci/benchmark_report.py:494-499` already accepts have no machine behind them. |

The envelope schema accepting a runner class is not the same as having a host that reports it. There is currently nothing to point `runs-on:` at.

## Current state after #211

`SCALING_THREAD_POINTS` is `(1, 2, 3, 4, 6, 8)` on the 4-vCPU `ubuntu-24.04` hosted runner: genuine per-core density from 1→4, with 6 and 8 kept as an explicitly labeled oversubscription band (<=2x logical cores). Honest, but the interesting part of an allocator scaling curve — where per-thread caches and cross-thread frees start to matter — lives past 4 threads, and we can only observe it under contention.

## What unblocks it

Either of:

1. **Move the repo under an organization** on a plan with larger runners, then add a runner group and point the `build-and-measure` job at e.g. `ubuntu-latest-8-cores`.
2. **Register a self-hosted runner** on a machine with stable topology and report it as `self-hosted-informational` (or `stable-host` once its noise profile is characterized).

## Work when unblocked

- Point `build-and-measure` in `.github/workflows/benchmark-scaling.yml` at the larger runner; keep the `informational` stamp.
- Re-derive `SCALING_THREAD_POINTS` against the new core count (dense to the core count, oversubscription band to ~2x). This changes the metric comparison key, so it starts a **new history lineage** — do not rewrite the existing one.
- `ci/check_benchmark_scaling_workflow.py` pins `runs-on: ubuntu-24.04` and a 20-minute `build-and-measure` budget; both need updating with the move, and the checker's cross-language thread-point control keeps the Rust and Python declarations in step.
[...]

Issue #211, opened 2026-08-14
feat(benchmarks): close the allocator-graph gap — plot the memory data we already collect, add Pareto/decay/fragmentation views, and densify scaling


zackees:
Parent phase: #183 (phase 6 advanced metrics)
Related measurement issues: #184 (6A RSS/fragmentation — **landed**), #185 (6B latency — landed), #187 (6D pprof tax — pending), #203/#208 (scaling cadence + density), #47 (per-OS graphs), #209 (overlay verification)
Design authority: #176 · Implementation guide: #177
PR count: one per work package below — **do not** land these in a single PR.

## Context

We already **measure** rich allocator memory behavior but barely **visualize** it. Issue #184 shipped `linux-process-memory-v1`, which captures — per fresh child, paired blocks, external `/proc/<pid>/smaps_rollup` sampling — a full **5 ms RSS timeline**, **post-drain RSS at 100 ms / 1 s / 5 s** (return-to-OS), and a **`fragmentation_proxy`** ratio (`sampled_peak_rss_delta / peak_live_requested`). All of it is validated by `ci/benchmark_report.py`.

But the only memory chart we render (`benchmark-memory.png`, `memory_png` at `ci/benchmark_report.py:2428`) draws a grid of **`sampled-peak-rss-bytes` bars** and discards everything else. Throughput is normalized to `upstream-mimalloc`; memory is shown in absolute bytes, un-paired. And thread scaling (`benchmark-scaling-*.svg`) is a hardcoded `SCALING_THREAD_POINTS = (1, 4, 16)` sweep (`:119`) on a 4-vCPU runner, so the 16-thread point is already 4× oversubscribed.

Against the field standard (mimalloc's own technical report, mimalloc-bench, micromalloc, rpmalloc, and the jemalloc/tcmalloc decay literature), a serious allocator memory comparison shows six views. We measure most of the underlying data and plot almost none of it:

| Canonical graph | Field reference | We measure? | We plot? |
|---|---|---|---|
| Paired peak-RSS **and** exec-time bars, **normalized to a baseline** | mimalloc TR | throughput + peak-RSS ✅ | throughput normalized ✅; RSS absolute/un-paired ⚠️ |
| **Speed–memory Pareto scatter** (X=overhead ratio, Y=ops/s) | micromalloc, rpmalloc | both axes ✅ | ❌ |
[...]

PR #218, opened 2026-08-14
feat(benchmarks): dense thread sweep, RSS-vs-threads, lineage-aware contract (#211 WP4)

zackees:
Work package 4 of issue #211, on top of WP1 (#213), WP2 (#214), WP3 (#217). This is the only WP that is not renderer-only.

## Changes

**Dense sweep.** `SCALING_THREAD_POINTS` goes `(1, 4, 16)` → `(1, 2, 3, 4, 6, 8)`: genuine per-core density 1→4 on the 4-vCPU hosted runner, with 6/8 kept as an explicitly labeled oversubscription band (≤2× logical cores). The old sweep's 16-thread point was already 4× oversubscribed with no intermediate points to interpolate from.

**RSS vs thread count.** A new optional `throughput-scaling-rss-v1` side-car: the runner samples `/proc/<pid>/smaps_rollup` externally every 5 ms per scaling child and keeps the peak over the measured block. Each `benchmark-scaling-*.svg` renders throughput and peak-RSS panels side by side when the side-car is present, and the single panel unchanged when it is not. Optional-by-construction so rows published before it existed stay valid history.

**Lineage-aware contract.** Changing the thread points changes the metric comparison key, so the sweep shape becomes a *lineage* rather than a constant: `SCALING_THREAD_POINT_LINEAGES` carries both shapes and every report is validated and drawn against the shape **it declares**, not the current one.

This one is not theoretical — I verified it against the live `benchmark-stats` branch, whose `latest.json` still declares `thread_points: [1, 4, 16]`:

```
# before the lineage change
ERROR: latest.json.scaling.thread_points: exactly [1, 2, 3, 4, 6, 8] required
# after
PASS rendered and sealed ...; manifest sha256=f1b23f70...   # byte-identical to main's render
```

Without it, the first scheduled run after this merged would have died on its own prior site.

**Cross-language control.** The Rust producer and the Python validator each declare the sweep, the block count, and both schema strings. Drift between them is silent until a scheduled run has already spent its budget and fails at overlay time. `ci/check_benchmark_scaling_workflow.py` now pins them to each other, with eight negative controls of its own (25 total, up from 17) — this is the "update the checker expectations and the `SCALING_THREAD_POINTS` contract together" item from the issue.

## Budget

Measured on the last real scaling run (`31761485183`): 326 s allocator build + 40 s suite build + 179 s sweep at 48 cells + 29 s overlay = 9 min. The sweep doubles to 96 cells → ~360 s, so ~13 min against the unchanged 20-minute `build-and-measure` budget. No timeout change needed.

## The 7k-line fixture

`rust/benchmark-suite/tests/fixtures/scaling/scaling-raw-run.json` is a **synthetic** fixture, not a captured run — `child_binary_sha256` is deliberately all-`9`s and the values are analytically shaped so the tests can assert exact coordinates. It grew from 144 to 288 samples purely because the cell count doubled (4 patterns × 6 thread points × 4 allocators × 3 blocks), plus the new per-sample `peak_rss_bytes`. Regenerated by the same fixture-spec builder that produced the original.

## Not included: the larger runner

WP4's third bullet asks scaling to move onto real cores. GitHub-hosted larger runners require an **organization** (`owner.type == User` here) and no self-hosted runner is registered (`total_count == 0`), so there is nothing to point `runs-on:` at. Escalated on #211 with the evidence and split out as #215.

## Notes

- Mixes `ci/` and `rust/` paths, which CLAUDE.md rule 2 permits — it forbids mixing **C-core** (`src/`, `include/`, `test/`, `CMakeLists.txt`) with `rust/`. No C-core path is touched.
- Changing the thread points starts a new history lineage; the existing sparse lineage is preserved, not rewritten.

## RED→GREEN

- A known RSS fixture must place every circle at its exact computed coordinate on the right-hand panel.
- A report without the side-car must render exactly one panel at `viewBox="0 0 1000 590"`.
- The previous `(1, 4, 16)` lineage must still validate **and** draw on its own axis (16 at the right edge, no 6).
- A shape belonging to no lineage (`[1, 5, 25]`) must be rejected.
- The checker must reject eight distinct Rust↔Python drifts.

Local: `pytest ci/tests` 72 passed, scaling policy selftest 25 controls, `ruff`/`pyright` clean, and the full `soldr cargo` test run for `benchmark-suite` green on Windows.


zackees:
**On the red checks** — all four failing jobs reproduce on `main` with controls; none are caused by this diff.

| job | failing thing | control on `main` |
|---|---|---|
| `test-win-gnu` | `ordinary_measured_region_performs_zero_harness_allocations` (`TinyFixed16`) | I re-ran main's own job `94677537031` with no code change: **byte-identical failure**. It was 6/6 green an hour earlier, so the trigger arrived from outside the repo. |
| `test (windows-latest)` | soldr broker `ErrorBackendSpawnFailed`, `could not compile xtask` | red on main (`31769950561`), survived 2 reruns |
| `test (macos-latest)` | `latency_sampling_buffers_do_not_allocate_after_warmup` | red on main, same test |
| `test (…)` ×7 in `cross` | `planted_serialized_control_is_rejected` (bench-harness) | red on main for 5+ runs, back to 2026-08-11 |
| `benchmark-sentinel (windows-latest)` | same soldr broker | red on **every** branch since at least 2026-08-11, including PRs that merged (#201, #207) |

Tracked in #219. None of these live in code this PR touches: the audit tests import `execution`/`scenarios`, `planted_control` is in `bench-harness`, and this diff changes only `scaling.rs`, `scaling_runner.rs`, a scaling fixture, and `ci/`. Locally the audit test passes 6/6 clean and 5/5 under 8 spinning CPU hogs on MSVC.

Everything that is green on `main` is green here: `c-unit` on ubuntu/windows-MSVC/windows-MinGW/macos (`MI_PPROF` ON and OFF), `python-lint`, the amalgamation check, and `rust-native` on ubuntu.

Issue #219, opened 2026-08-14
ci: rust-native is red on main (soldr broker on windows, latency-audit flake on macos) — a required merge gate is unusable


zackees:
CLAUDE.md rule 3 makes `rust-native` a merge gate for every PR. It is currently failing on `main` itself, so the gate cannot be satisfied by any PR and provides no signal.

## Evidence

Every recent `rust-native` run on `main` is red:

```
$ gh run list --workflow rust-native.yml --limit 8 --json conclusion,createdAt,headBranch
2026-08-14T04:28 failure main
2026-08-14T04:12 failure main
2026-08-14T04:03 failure main
2026-08-14T03:54 failure main
2026-08-14T02:22 failure main
2026-08-14T01:43 failure main
```

Per-job on run `31769950561` (main):

| job | result |
|---|---|
| `test (ubuntu-latest)` | success |
| `test-win-gnu` | success |
| `test (windows-latest)` | **failure** |
| `test (macos-latest)` | **failure** |

## Two distinct causes

**1. Windows — the soldr broker never comes up.** The MSVC job dies before any test runs:

```
soldr: broker SESSION compile failed: soldr broker is unreachable: SESSION compile failed
  (output_started=false): The system cannot find the file specified. (os error 2)
soldr: broker SESSION compile failed: soldr broker could not provide daemon route
  soldr-daemon-390cc0499331c235c258f5f104f17007 within 30000ms: ... dial v2 broker pipe at
  "\.\pipe\rpb-v2-soldr-daemon-715f42bbb0152cfe-0": The system cannot find the file specified.
  (os error 2) ... code: ErrorBackendSpawnFailed
error: could not compile `xtask` (bin "xtask")
```

Preceded by `Cached Rust toolchain failed validation; reinstalling 1.94.1` and `cook: failed with exit 101 after 35.7s; continuing without cooked deps`, which suggests the setup-soldr cache/daemon handshake, not the repo. It reproduces across reruns (I retried twice on #218) and also takes down `benchmark-sentinel (windows-latest)`, which has been red on every branch since at least 2026-08-11.

Note `test-win-gnu` passes, so this is specific to the MSVC job's soldr daemon, not to Windows generally.

**2. macOS — an allocation-audit test fails.**

```
thread 'latency_sampling_buffers_do_not_allocate_after_warmup' (21985) panicked at
  benchmark-suite/tests/timed_allocation_audit.rs:228:13
```

## Why this matters

[...]

PR #224, opened 2026-08-20
fix(rust): use C11 atomics for Windows ARM64

### LINKED ISSUE #223: fix(rust): use C11 atomics for Windows ARM64 clang-cl
## Context

`mimalloc-pprof 0.9.2` fails when cross-compiled for `aarch64-pc-windows-msvc` through xwin/clang. The vendored mimalloc v3 source selects MSVC ARM64 atomics that clang does not expose while the amalgamation is compiled as C.

A zccache release build reproduced the failure in zackees/zccache#864. Forcing `-TP` proves that the amalgamation builds through mimalloc's supported C++ atomics path, but a workspace-global `CFLAGS=-TP` is unsafe: it also compiles unrelated native C dependencies such as `ring` as C++ and breaks them.

The current `rust/mimalloc-pprof/build.rs` has no target-specific handling, so consumers cannot apply the necessary flag to this crate alone.

## Proposal

In `rust/mimalloc-pprof/build.rs`, detect the exact `aarch64-pc-windows-msvc` target and add the compiler's C++-language flag to this crate's `cc::Build` only. Keep all other targets and dependencies in C mode.

Add a focused cross-build regression in CI for the published Rust crate on Windows ARM64, ideally through the same clang/xwin route that exposed the bug.

## Acceptance criteria

- A focused RED repro demonstrates that the unmodified crate fails for `aarch64-pc-windows-msvc` under clang/xwin.
- The package-scoped build-script change turns that repro GREEN.
- The fix does not require consumers to set global `CFLAGS` or `CXXFLAGS`.
- Existing native Rust and amalgamation checks remain green.
- A consumer graph containing both `mimalloc-pprof` and a native C dependency such as `ring` can cross-build without forcing the latter into C++ mode.

## Decisions

- **Severity: P1 release blocker.** This currently blocks zccache's Windows ARM64 release artifact.
- **Fix location: crate build script.** Only the crate knows that its vendored source supports/needs the C++ atomics path; consumer-wide flags are too broad.
- **Scope: exact ARM64 MSVC target.** Other targets already build as C and should remain unchanged.
- **Validation: clang/xwin cross-build.** That is the failing production toolchain and catches the package-scoping regression.

## Related issues

- zackees/zccache#864 tracks the blocked downstream release workflow.

zackees:
## Summary
- add an explicit C11-atomics escape hatch to mimalloc's atomic abstraction
- select that path only for `aarch64-pc-windows-msvc` in the Rust crate build script, keeping the amalgamation in C mode
- avoid both missing ARM64 clang-cl intrinsics and consumer-wide C/C++ flags
- validate the exact ARM64 Windows target in the cross-build matrix

## Validation
- hosted first attempt reproduced why the C++ path is invalid for the current v3 amalgamation
- `soldr cargo test -p mimalloc-pprof --test build_target` (2 passed)
- `soldr cargo test -p mimalloc-pprof --locked` (all passed)
- `soldr cargo clippy -p mimalloc-pprof --lib --test build_target --locked -- -D warnings`
- `soldr cargo fmt -p mimalloc-pprof -- --check`
- `soldr cargo run -p xtask -- check`

Closes #223

Downstream: zackees/zccache#864

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -109,6 +109,7 @@ jobs:
       matrix:
         target:
           - x86_64-pc-windows-msvc
+          - aarch64-pc-windows-msvc
           - x86_64-apple-darwin
           - aarch64-apple-darwin
     defaults:
--- a/include/mimalloc/atomic.h
+++ b/include/mimalloc/atomic.h
@@ -28,7 +28,20 @@ terms of the MIT license. A copy of the license can be found in the file
 // instead of passing the memory order as a parameter.
 // -----------------------------------------------------------------------------------------------
 
-#if defined(__cplusplus)
+#if defined(MI_USE_C11_ATOMICS)
+// Some MSVC-compatible C compilers provide C11 atomics but not every
+// architecture-specific intrinsic used by the plain MSVC wrapper below.
+#include <stdatomic.h>
+#define  mi_atomic(name)          atomic_##name
+#define  mi_memory_order(name)    memory_order_##name
+#if (__STDC_VERSION__ >= 201710L) // c17, see issue #735
+ #define MI_ATOMIC_VAR_INIT(x)    x
+#elif !defined(ATOMIC_VAR_INIT)
+ #define MI_ATOMIC_VAR_INIT(x)    x
+#else
+ #define MI_ATOMIC_VAR_INIT(x)    ATOMIC_VAR_INIT(x)
+#endif
+#elif defined(__cplusplus)
 // Use C++ atomics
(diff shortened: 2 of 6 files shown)


zackees:
Merge basis: the exact target gate uild (aarch64-pc-windows-msvc) passes, as do every native C gate and supported Ubuntu/Windows Rust test. The macOS Rust allocation-audit failure remains the pre-existing behavior tracked by #219; the downstream cross 	est jobs fail in the existing runner/test-binary harness after their corresponding build jobs succeed.

Issue #194, opened 2026-08-11
handoff: complete benchmark dashboard publication pipeline and close #177


zackees:
## Context

This is a handoff issue from the session that implemented Phases 3-5A of the public benchmark dashboard pipeline (#177). Three phases were merged (PRs #191, #192, #193), infrastructure fixes were pushed to main, and the Phase 5A live publication workflow is currently executing (run #31447674464).

### Phase 3 (#180) — Merged PR #191
- `rust/benchmark-suite/src/model.rs`: typed publication envelope, LatestReport, HistoryRow, RawSample, child protocol types
- `rust/benchmark-suite/src/stats.rs`: Type-7 quantiles, paired bootstrap statistics (10k resamples, splitmix64-rejection)
- `rust/benchmark-suite/src/validate.rs`: fail-closed semantic validation of 4-allocator 30-cell matrices (≥15 blocks)
- `rust/benchmark-suite/src/comparison_key.rs`: deterministic SHA-256 compatibility key
- `rust/benchmark-suite/src/report.rs`: builds LatestReport + HistoryRow from validated data
- `rust/benchmark-suite/src/bin/benchmark-validate.rs`: CLI validator with selftest
- 3 JSON schemas (raw-run, latest, history) with `additionalProperties: false`
- `ci/benchmark_report.py`: Python renderer (PNG, HTML, manifest, detached digest, site validation)
- 25 Rust tests + 9 Python tests

### Phase 4 (#181) — Merged PR #192
- `.github/workflows/benchmark-stats.yml`: initial workflow_dispatch-only dry run, contents:read
- `ci/check_benchmark_workflow.py`: structural policy checker with selftest
- `ci/tests/test_benchmark_workflow.py`: 13 policy tests

### Phase 5A (#182) — Merged PR #193
- Added daily schedule (09:17 UTC), publication eligibility gate, 6 jobs:
  build-and-measure → artifact-audit → publish-branch → package-pages → deploy-pages → publication-audit
- History carry-forward: git ls-remote benchmark-stats branch, fail-closed on errors
- publish-branch: contents:write, force-with-lease against prior SHA, create-only first run
- deploy-pages: pages:write + id-token:write, github-pages environment
- Policy checker extended for Phase 5 permissions model

### Infrastructure fixes (pushed directly to main)
- Fixed YAML parse errors (multi-line commit message with unquoted colons)
- Switched actions/* to immutable semver tags (SHAs not always reachable from default branch)
[...]

PR #229, opened 2026-08-23
docs(design): allocator shared-library cache (#194)

zackees:
Design document for #194 — cache compiled allocator artifacts between CI runs.

**Draft.** This was written 2026-08-10 and sat **untracked** in a working tree since; it existed in exactly one place and nothing referenced it. Opening it as a draft makes it durable and reviewable. It has not been re-reviewed against current `main`.

## The problem it addresses

`benchmark-stats.yml` builds four native allocator libraries from source on **every** run — tcmalloc (Bazel), jemalloc (autoconf-make), upstream-mimalloc and mimalloc-pprof (cmake-ninja). That is **20–30 minutes** in `build-and-measure` against a 50-minute timeout, and it is the dominant cost of a benchmark run. Nothing changes between runs: same pinned commits, same patches, same toolchain.

It also names a second problem found along the way: adding a competitive allocator today requires touching five scattered dispatch points (lockfile, C adapter, three spots in the Python builder, `build.rs`'s match, and the Rust adapter allowlist). The interfaces are correct; the dispatch is not.

## Proposed shape

An `allocator-lock.json` describing each allocator's source pin, patches, build system and flags; each built as a shared library exporting a stable versioned ABI that the harness dlopens; and a content-addressed cache keyed on the input hash — no timestamp heuristics, no manual invalidation.

Stated goal is that adding an allocator becomes one lockfile entry plus one adapter C file, with zero dispatch-table edits.

Existing security invariants are explicitly preserved: every archive checksummed before extraction, every symbol verified post-link, and no second allocator contaminating a child process.

## Status

Document only — no implementation. Sections cover the ABI, build, cache, dynamic loading, migration plan, resolved design questions, and a validation strategy. Review of whether the approach is still the one we want is the next step; that is why this is a draft.

### DIFF
--- a/design/allocator-so-cache.md
+++ b/design/allocator-so-cache.md
@@ -1,0 +1,800 @@
+# Design: Allocator Shared-Library Cache
+
+**Status:** Draft
+**Date:** 2026-08-10
+**Issue:** [#194](https://github.com/zackees/mimalloc-pprof/issues/194)
+
+## Status — revised 2026-09-06
+
+Written 2026-08-10 against a pipeline that has since changed. It was re-checked against
+`main` before landing, and **two of its premises did not survive that check**. The document
+is kept because its second problem is real and larger than it claimed; it is not an
+approved plan, and the priority order below supersedes the Goals section.
+
+**Premise 1 — "20–30 minutes, the dominant cost" — is stale.** Measured on
+[run 34034422851](https://github.com/zackees/mimalloc-pprof/actions/runs/34034422851)
+(benchmark-stats, 2026-09-06), the `build native allocator libraries` step ran
+**6 m 15 s** inside a `build-and-measure` job that took **48 m 36 s** — about 13 %, and the
+measurement phase is now the dominant cost, as it should be. Caching the allocator build is
+therefore worth roughly six minutes a run, not twenty-five. That is worth having, but it no
+longer justifies a shared-library/dlopen redesign on its own.
+
+**Premise 2 — "adding an allocator touches five places" — understates it.** The natural
+experiment is #325, the only allocator actually added since: `oven-sh/mimalloc` as a fifth
+row touched **48 files, +3366/−472**, spread across `config.rs`, `latency.rs`,
+`latency_runner.rs`, `memory.rs`, `memory_runner.rs`, `model.rs`, `orchestration.rs`,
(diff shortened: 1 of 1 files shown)


zackees:
Re-checked against `main` and rebased (the branch was 523 commits behind). **Two premises did not survive the check**, so the revision is recorded at the top of the document rather than landing claims that no longer hold.

**"20–30 minutes, the dominant cost" is stale.** On [benchmark-stats run 34034422851](https://github.com/zackees/mimalloc-pprof/actions/runs/34034422851) (2026-09-06):

| step | duration |
|---|---|
| `build native allocator libraries` | **6 m 15 s** |
| `build-and-measure` (whole job) | **48 m 36 s** |

~13 %, with measurement now dominant — as it should be. Caching is worth about six minutes a run, not twenty-five, which is no longer enough on its own to justify the shared-library/dlopen redesign.

**"Adding an allocator touches five places" understates it.** The natural experiment is #325, the only allocator actually added since: **48 files, +3366/−472**, across `config.rs`, `latency.rs`, `memory.rs`, `model.rs`, `orchestration.rs`, `provenance.rs`, `runner.rs`, `scaling.rs`, `validate.rs`, eight test files and the fixtures. The five dispatch points it names are real and still in place (`ADAPTER_SYMBOLS`/`COMPETITOR_SYMBOLS` at `ci/build_benchmark_allocators.py:59,69`, `adapter_include_directories` `:668`, `find_primary_library` `:810`, `compile_adapter` at `build.rs:245`) — but they are a minority of the real cost, most of which is allocator-count assumptions in the statistics and validation code that no dispatch-table change reaches.

**Already landed since it was written:** `allocator-lock.json` is exactly the lockfile this proposes, with pins, archive SHA-256s, patches, build flags and a `--selftest` — as *static* archives, not shared libraries. Its `upstream-mimalloc` pin moved to `6def7be9` in #332. There is still no `actions/cache` in any benchmark workflow.

**Revised priority in the doc** (supersedes its Goals): (1) measure how much of #325's diff a lockfile-plus-adapter shape would actually have removed before designing for it; (2) caching on its own merits, as an `actions/cache` keyed on the lockfile hash — needs none of the ABI work; (3) dlopen last, if at all.

**Ordering caveat**, also in the doc: `benchmark-scaling.yml` has not published since 2026-08-14 (#376, tracked in #377). Adding machinery to a pipeline that cannot publish would be hard to validate.

Marked ready for review. Landing it makes the document durable — its original purpose — and is **not** approval to implement.

PR #225, opened 2026-08-20
chore(release): bump mimalloc-pprof to 0.9.3

zackees:
## Summary
- bump the Rust crate and lockfile to 0.9.3
- document the Windows ARM64 build fix from issue #223

## Validation
- soldr cargo check -p mimalloc-pprof

PR #231, opened 2026-08-23
fix(atomic): detect C11 atomics instead of allowlisting aarch64-pc-windows-msvc

### LINKED ISSUE #223: fix(rust): use C11 atomics for Windows ARM64 clang-cl

### LINKED ISSUE #230: Detect C11 atomics instead of allowlisting aarch64-pc-windows-msvc
## Context

`include/mimalloc/atomic.h` picks an atomics backend with this chain:

```c
#if   defined(MI_USE_C11_ATOMICS)   // 1. explicit override  -> <stdatomic.h>
#elif defined(__cplusplus)          // 2. C++                -> <atomic>
#elif defined(_MSC_VER)             // 3. MSVC C wrapper     -> mi_atomic_* / Interlocked
#else                               // 4. everything else    -> <stdatomic.h>
#endif
```

Branch 3 tests `_MSC_VER`, which means *"claims MSVC source compatibility."* What it actually needs to select is *"lacks C11 atomics."* Those are not the same set, and `clang-cl` is the gap: it defines `_MSC_VER` for compatibility but is clang, and has complete C11 atomics.

So `clang-cl` takes branch 3, gets the hand-rolled MSVC wrapper, and on ARM64 hits `__ldar64` / `__stlr64` — Interlocked intrinsics clang-cl does not declare. That is the failure #223 fixed.

#223 shipped the fix in 0.9.3 as a build-script define, deliberately scoped to one triple:

```rust
// rust/mimalloc-pprof/build_target.rs, in full
pub(crate) fn needs_c11_atomics(target: Option<&str>) -> bool {
    target == Some("aarch64-pc-windows-msvc")
}
```

That was the right call under #223's constraints — it was a P1 release blocker and the narrow scope was the low-risk move. Its Decisions log records the reasoning: *"Scope: exact ARM64 MSVC target. Other targets already build as C and should remain unchanged."*

This issue revisits that decision now that the release pressure is gone, because the allowlist is standing in for a capability test.

## The problem with the allowlist

- **It is a target list approximating a compiler property.** `x86_64-pc-windows-msvc` under clang-cl also takes the MSVC wrapper; it only survives because x64 happens to have the intrinsics ARM64 lacks. Nothing records that, so the next architecture (arm64ec, or a future clang-cl target) fails the same way and looks like a new bug.
- **It only helps consumers who go through the Rust build script.** The CMake build, `src/static.c` consumers, and anyone vendoring the C directly still hit branch 3 with clang-cl. The bug lives in the C, but the fix lives in Rust.
- **Only downstream discovers it.** zackees/zccache#1439 spent a release cycle on a global `CFLAGS=-TP` workaround for this, which then compiled `ring`'s C sources as C++ and broke the 1.13.6 release on a different axis.

## Proposal

Feature-detect instead of enumerating targets. C11 provides the exact test — `__STDC_NO_ATOMICS__`, defined precisely when `<stdatomic.h>` is unavailable. Insert a detection branch ahead of the MSVC one:

```c
#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__)
// C11 atomics: any compiler that has them, including clang-cl, which defines
// _MSC_VER for source compatibility but is not MSVC.
#elif defined(_MSC_VER)
// MSVC C wrapper, now reached only by compilers that genuinely lack C11 atomics.
```

`MI_USE_C11_ATOMICS` stays as an explicit override. Branch 4 is unchanged.

### Why this is safe for real MSVC

The guard is false for every `cl.exe` configuration, so branch 3 keeps its current population:

| Compiler | `__STDC_VERSION__` | `__STDC_NO_ATOMICS__` | Branch taken |
|---|---|---|---|
| `cl.exe`, no `/std:c11` | **not defined** | — | 3 (wrapper) — unchanged |
| `cl.exe` `/std:c11`, no `/experimental:c11atomics` | 201112L | **defined** | 3 (wrapper) — unchanged |
| `cl.exe` `/std:c11 /experimental:c11atomics` | 201112L | not defined | new C11 branch |
| `clang-cl` (default C17) | 201710L | not defined | **new C11 branch — the fix** |

Only the third row changes behavior for `cl.exe`, and only when the user has explicitly opted into MSVC's C11 atomics.

If reviewers want zero behavior change for `cl.exe` under any flags, the conservative variant is to add `defined(__clang__) &&` to the new guard. That narrows it to clang-family compilers at the cost of not being a portable capability test — worth deciding on this issue rather than in review.

## Consequences

- `needs_c11_atomics` and the `MI_USE_C11_ATOMICS` define in `build.rs` become dead and should be removed; the C detects it. This is a behavior-preserving removal — same backend selected, chosen one layer down.
- CMake, `static.c`, and direct-vendoring consumers get the fix they currently do not get.
- Upstreamable to microsoft/mimalloc as a few guarded lines under rule 6. Upstream has the identical chain and no `__STDC_NO_ATOMICS__` test anywhere, so it has the same latent bug.

## Acceptance criteria

- RED: the crate built for `aarch64-pc-windows-msvc` through clang/xwin **with `needs_c11_atomics` removed and no C change** still fails on the missing ARM64 intrinsics.
- GREEN: the same build succeeds with the `atomic.h` change and no build-script define.
- `cl.exe` (MSVC) and MinGW `c-unit` jobs stay green with `MI_PPROF=ON` and `OFF` — confirming branch 3's population is unchanged for real MSVC.
- `rust-native` green; amalgamation drift check green (the C change must be carried into `rust/mimalloc-pprof/vendor/` as a separate commit per rule 2).
- A consumer graph with both `mimalloc-pprof` and `ring` cross-builds for Windows ARM64 with no global `CFLAGS`/`CXXFLAGS`.

## Related

- #223 — the 0.9.3 fix this supersedes; its Decisions log is the reason the current scope is narrow.
- zackees/zccache#1439 — the downstream `-TP` workaround that broke `ring`.
- zackees/zccache#1491 — downstream contract preventing the fork-pin that this issue's fix makes unnecessary.

zackees:
Closes #230.

## What

`include/mimalloc/atomic.h` selected the hand-rolled MSVC atomics wrapper on `defined(_MSC_VER)`. That macro means *"claims MSVC source compatibility"*, not *"lacks C11 atomics"* — and clang-cl defines it while being clang, with full C11 atomics. So clang-cl took the wrapper and reached for `__ldar64`/`__stlr64`, Interlocked intrinsics it does not declare on ARM64. That is the `aarch64-pc-windows-msvc` failure from #223.

The wrapper is now gated on a capability macro derived from the standard test:

```c
#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) && !defined(__STDC_NO_ATOMICS__)
#define MI_HAS_C11_ATOMICS 1
```

**Both** condition chains move together — the backend selection at the top of the file *and* the typed-ptr/Interlocked split further down. Changing only one would give clang-cl C11 atomics in one chain and the Interlocked implementation in the other.

## Why this supersedes the 0.9.3 fix

#223 shipped a build-script allowlist, which was the right call under release pressure. Its limits:

- **It is a target list approximating a compiler property.** `x86_64-pc-windows-msvc` takes the same wrong branch under clang-cl and only survives because x64 has the intrinsics ARM64 lacks. The next architecture fails the same way and looks like a new bug.
- **It only reached consumers who build through the build script.** CMake, `static.c`, and direct-vendoring consumers still hit the bug. The defect was in the C; the fix was in Rust.

So the allowlist and its `MI_USE_C11_ATOMICS` define are removed here — behavior-preserving, same backend chosen one layer down. `MI_USE_C11_ATOMICS` still works as an explicit override, and `MI_HAS_C11_ATOMICS` can be defined to 0/1 to override detection.

## Real MSVC is unaffected

The guard is false for every `cl.exe` configuration except one the user opts into explicitly:

| Compiler | `__STDC_VERSION__` | `__STDC_NO_ATOMICS__` | Branch |
|---|---|---|---|
| `cl.exe`, no `/std:c11` | not defined | — | wrapper — unchanged |
| `cl.exe` `/std:c11` | 201112L | **defined** | wrapper — unchanged |
| `cl.exe` `/std:c11 /experimental:c11atomics` | 201112L | not defined | C11 |
| `clang-cl` (default C17) | 201710L | not defined | **C11 — the fix** |

## Verification

- Preprocessed the **real** guards (extracted from the file, not retyped) under all six configurations above plus the explicit override. Every case matches the table, and the two chains agree in all six.
- Amalgamation regenerated with `xtask amalgamate-c`; drift check green.
- Crate builds — the amalgamation compiles for `x86_64-pc-windows-msvc`, which exercises the unchanged real-MSVC wrapper path.
- Full `mimalloc-pprof` suite green. (The one `unused import: prof` warning is pre-existing on `main`; confirmed by stashing.)

**Not verified locally:** the actual `aarch64-pc-windows-msvc` clang/xwin cross-build — no clang-cl, cargo-xwin, or that target on this machine. That is #230's RED/GREEN criterion and needs the CI gates, which is what this PR is for. Please don't merge on the local evidence alone.

## Commits

Split per repo rule 2 — C-core and `rust/` never in one commit:

1. `fix(atomic):` — `include/mimalloc/atomic.h` only
2. `refactor(rust):` — regenerated amalgamation, allowlist removal, contract tests

## Follow-up

Version bump to 0.9.4 as its own PR, per the #225 precedent, once this is green. Downstream zackees/zccache#1439 can then drop its pin entirely.

Upstreamable: microsoft/mimalloc has the identical chain and no `__STDC_NO_ATOMICS__` test anywhere, so it carries the same latent bug — a few guarded lines under rule 6.


zackees:
Merging via **rebase**, not squash, deliberately: squashing would collapse the C-core and `rust/` commits into one commit touching both `include/` and `rust/`, which is exactly what repo rule 2 forbids ("keeps upstream cherry-picks clean"). Rebase preserves the split and keeps `main` linear.

**Merge gates (rule 3) — all green:**
- `c-unit` ubuntu / windows-MSVC / windows-MinGW / macOS with `MI_PPROF=ON` ✅
- the `OFF` job (`ctest-pprof-off`) ✅
- `rust-native` ✅
- plus `build (aarch64-pc-windows-msvc)` ✅ — #230's GREEN criterion, proving the C11 capability detection replaces the removed build-script allowlist
- `vendored C amalgamation matches src/` ✅, `address sanitizer` ✅, `fuzz` ✅, `memory-gate` ✅

**Known red, pre-existing and unrelated:** `cross.yml`'s `test (<triple>)` matrix. It has failed on every `main` run since at least 2026-08-14, including run 32456311287 at `6fcb1966` — this branch's exact base — where six of these jobs failed. Root cause is the "Run every staged test binary" step globbing the artifact dir and running a bench-harness test that spawns a benchmark executable the build job never stages:

```
valid normal benchmark: Spawn(Os { code: 2, kind: NotFound, message: "No such file or directory" })
##[error]Process completed with exit code 101
```

Nothing to do with atomics. Worth its own issue.

Issue #128, opened 2026-08-01
Meta: implement the imports worth taking from the upstream branch sweep


zackees:
Action items from the upstream branch sweep. Background and full evidence: **#115** (48 findings, 43 refs classified) and **#84** (the sweep's own specification). Those are closed; this issue is the only live thread.

The sweep's headline is a negative result and it shapes this plan: **almost nothing is worth cherry-picking. The pin bump is how upstream work reaches us.** Of 48 findings, 15 were actionable and most of those are our-own-code work informed by upstream prior art, not imports.

---

## A. Arrives free with the next pin bump — do not cherry-pick

`upstream/dev3` is 6 commits past our pin `bcee5a88e`. Two are real; both come along automatically when the pin moves. Cherry-picking them now would only create merge friction later.

- [x] **A1 (marked in-tree for the pin bump, #146) — `b0ac42ebc` singleton page rollback.** `mi_arenas_page_singleton_alloc`'s `_mi_page_init` failure path calls raw `_mi_arenas_free` where both sibling paths call `_mi_arenas_page_free`, skipping `_mi_page_map_unregister`, the arena pages-bitmap clear, and the page-stat decrements — and passing the wrong pointer under `MI_PAGE_META_IS_SEPARATED`. **Latent, not live:** at our pin `commit=true` ⇒ `slice_committed==0` ⇒ the only `return false` in `mi_page_extend_free` is unreachable for singletons. Was #120.
- [x] **A2 (marked in-tree for the pin bump, #146) — `1fb345674` / `f9a74121f` `pages_full_size` uses `capacity` instead of `reserved`.** Cosmetic at our pin: `pages_full_size` is write-only — nothing reads it, so a wrapped value has zero release-build effect. Take it with the bump, don't chase it.

**Action: fold into the next pin-bump PR (#80's successor). No separate work.**

---

## B. Real defects in our pin — port from `dev3-meta`

These need actual porting: `dev3-meta` carries the `theap_meta` architecture our pin lacks, so none is a clean cherry-pick.

[...]

Issue #83, opened 2026-07-31
realloc and zeroing-realloc: already in C, missing from Rust and unverified in the profiler


zackees:
Reported as: "see if mimalloc supports realloc and if not add support for this, also a zeroing version of realloc."

**Answer to the question as asked: both already exist in C.** Checked before filing, so this issue is about the gaps that actually remain rather than the feature request.

## What already exists (`include/mimalloc.h`)

**Plain realloc:**

| API | Line |
|---|---|
| `mi_realloc(p, newsize)` | 111 |
| `mi_expand(p, newsize)` | 112 |
| `mi_reallocn(p, count, size)` | 130 |
| `mi_realloc_aligned(p, newsize, alignment)` | 152 |
| `mi_realloc_aligned_at(p, newsize, alignment, offset)` | 153 |

**Zeroing realloc — this exists too:**

| API | Line |
|---|---|
| `mi_rezalloc(p, newsize)` | 276 |
| `mi_recalloc(p, newcount, size)` | 277 |
| `mi_rezalloc_aligned(p, newsize, alignment)` | 279 |
| `mi_rezalloc_aligned_at(...)` | 280 |
| `mi_recalloc_aligned(...)` | 281 |
| `mi_recalloc_aligned_at(...)` | 282 |

They zero the *newly grown* region, which is the semantic people usually want and that plain `realloc` does not provide. Covered by `test/test-api-fill.c` and `test/test-api.c`.

**So there is nothing to add on the C side.** Closing this as "already supported" would be defensible — except for the following.

### 1. The Rust crate does not expose the zeroing realloc family at all

`grep -rn "rezalloc\|recalloc" rust/ --include=*.rs` returns **0 hits**.

The crate exposes `alloc_zeroed` (`rust/mimalloc-pprof/src/lib.rs:38`) and routes `GlobalAlloc::realloc` to `mi_realloc_aligned` (line 35), but `mi_rezalloc` / `mi_recalloc` and their aligned variants are not reachable from Rust — not through the sys crate, not through a safe wrapper.

That matters because `GlobalAlloc` has no `grow_zeroed`; a Rust caller who wants "grow this buffer and zero the new tail" currently has to grow and then `memset` themselves, doing work mimalloc would have done more cheaply — and, with #67's zero-tracking, possibly for free.

[...]

Issue #78, opened 2026-07-31
Investigate the Bun mimalloc fork end-to-end and ingest everything useful, with a sourced feature table in the README


zackees:
Bun (`oven-sh/bun`) vendors mimalloc and maintains a real fork of it at
[`oven-sh/mimalloc`](https://github.com/oven-sh/mimalloc). It is the most actively
maintained v3-line fork found so far, it is rebased on a **newer** upstream base than we
are, and it is exercised under a production JavaScript runtime — a workload profile our
test suite does not resemble.

We have only ever looked at **two** hunks from it, both in #66, and both turned out to be
`upstream-fixed-differently`. That was a narrow, issue-scoped look. This issue is the
opposite: a systematic sweep of the whole fork.

**Do not start this before #66's pin decision is acted on.** Importing against a stale
base is how forks become unrebasable, and #66 already recommends bumping to the dev3 tip.

## Known coordinates (from #66, verified)

| | |
|---|---|
| Repo | `github.com/oven-sh/mimalloc` |
| Default branch | `bun-dev3-v2` (HEAD `d078ad06` at time of writing) |
| Merge-base with `upstream/dev3` | `ce2c584c` — 20 commits newer than our pin `579f8c0e`, 72 behind upstream tip |
| Their test hook | `mi_debug_fail_os_commit_after` — **we do not have this**, so importing their test corpus is a core-file change under repo rules 2 and 6 |

## Scope

1. **Enumerate every divergence.** `git diff ce2c584c..bun/bun-dev3-v2` over `src/` and
   `include/`, plus their full commit log. Produce a complete inventory, not a sample —
   the point of this issue is that we have only spot-checked so far.

2. **Classify every hunk** with the #66 taxonomy — `already-upstream` /
   `upstream-fixed-differently` / `fork-only` — each with the upstream SHA cited for the
   first two. Two of two candidates classified as "drop" last time; expect a high drop
   rate again and treat that as a finding, not a failure.

3. **For each `fork-only` hunk, decide import / skip** with a reason. Per #66's C.2 rule,
   **import only with a test that fails without it.** If Bun's own reproducer depends on
   `mi_debug_fail_os_commit_after`, decide separately whether to port that hook — it is a
   core-file change and needs its own justification.

4. **Look beyond bug fixes**, which is what the narrow #66 pass missed by construction:
   - allocation-hot-path or fast-path changes and any benchmark numbers they published
[...]

PR #233, opened 2026-08-23
docs: 0.9.x release history in the README, and the missing 0.9.0 changelog entry

zackees:
## What this adds

A **Release history** section in the root `README.md`, placed immediately after
"Choosing a version: v2 or v3" — the reader picks a line, then sees what happened on it.
One table row per published 0.9.x version (0.9.3, 0.9.2, 0.9.1, 0.9.0) with the
crates.io publish date and a summary of what landed, followed by a short note on the
unreleased 0.9.4 work already merged to `main`.

## On duplication: summary in the README, record in the CHANGELOG

`rust/mimalloc-pprof/CHANGELOG.md` already carries good long-form notes. Restating them
in the README guarantees the two drift apart the first time someone updates one and not
the other, so the README section is deliberately a **table plus a link**: 1–3 lines per
version, and an explicit sentence saying the CHANGELOG is the record. A table rather
than prose because the README is already table-heavy (the version-choice matrix, the
bug tables, the benchmark tables) and it matches that voice.

The consequence is that the link target has to be complete, which is why this PR also
touches the CHANGELOG — see below.

**Two commits, strictly separated**, per the repo rule against mixing paths:

| Commit | Paths |
|---|---|
| `docs(rust): add the missing 0.9.0 changelog entry, complete 0.9.3` | `rust/mimalloc-pprof/CHANGELOG.md` only |
| `docs: add a 0.9.x release history to the README` | root `README.md` only |

No C-core path (`src/`, `include/`, `test/`, `CMakeLists.txt`) is touched. No version is
bumped and no `0.9.4`/Unreleased section is added to the CHANGELOG — the bump PR owns
that entry.

## Two gaps found while deriving this

**1. The CHANGELOG had no 0.9.0 entry.** It jumps from 0.9.1 to 0.8.0, whose entry still
reads "0.9.0 is reserved for a v3 mimalloc port (issue #29, in progress)". 0.9.0 shipped
2026-07-31 and is the release that put the v3 line on crates.io. Reconstructed from the
`v0.8.0..v0.9.0` range and the issues its commits cite: the v3 port of the profiler and
memory-events (#29, #44), v3 promoted to mainline with v2 preserved on the `v2` branch
(#46), v3 allocator counters exposed as `ProfStats::heap` (#43), the MinGW thread-exit
leak and the `mi_heap_new`/`mi_subproc_new` bootstrap crash (both `#43`-tagged commits;
the README's "Bugs fixed in older versions" independently records both as fixed in
0.9.0), and the new degenerate-pattern/leak-guard/aligned-alloc tests (#49, #51).

**2. 0.9.3 is more than the ARM64 atomics fix.** The CHANGELOG lists only #223. But
#129 ("Migrate the v4 line into v3 (main)") merged after 0.9.2 shipped and is an
ancestor of the `v0.9.3` tag, so 0.9.3 is also the first release carrying: the upstream
engine pin `579f8c0e` → `bcee5a88` (#80, #113), `mi_option_purge_zeroes` (#67, #79),
the zeroing-`realloc` family from Rust (#83, #108), and several allocator fixes
(#128, #78). `mi_option_purge_zeroes` is new public API that the 0.9.1 notes explicitly
deferred to v4, so omitting it understates the release.

Verified rather than assumed:

```
git merge-base --is-ancestor eec74891 v0.9.3   # yes
git merge-base --is-ancestor eec74891 v0.9.2   # no
git show v0.9.3:rust/mimalloc-pprof/vendor/mimalloc-pprof-amalgamated.h | grep purge_zeroes
git show v0.9.3:rust/mimalloc-pprof/src/lib.rs | grep rezalloc
```

## Sources

- `git log v0.8.0..v0.9.0`, `v0.9.2..v0.9.3`, and the release-bump commit messages
  (`4c140b68`, `0d367832`, `6ad84e4d`, `49785a4d`), whose bodies are unusually detailed.
- `rust/mimalloc-pprof/CHANGELOG.md` for 0.9.1–0.9.3.
- crates.io API for publish dates — parsed structurally, **UTC**, not commit dates:
  0.9.0 `2026-07-31`, 0.9.1 `2026-08-01`, 0.9.2 `2026-08-01`, 0.9.3 `2026-08-21`.
  0.9.1/0.9.2 were committed Jul 31 local (`-0700`); the README says the dates are
  crates.io publish dates so the two don't look self-contradictory.
- `gh issue view` / `gh pr view` on **every** number cited: 29, 43, 44, 46, 49, 51, 62,
  67, 68, 72, 78, 79, 80, 83, 91, 108, 113, 128, 129, 223, 224, 230, 231.
- For 0.9.4: commits `35a490f6` (C-core) + `be3e621f` (`rust/`) from #231 / issue #230.

## Left out, or hedged

- **`Cross-compilation` is untouched.** It still describes the 0.9.3 build-script
  allowlist, which is what is actually on crates.io today. Rewriting it now would
  document an unreleased mechanism. The 0.9.4 note points at it and says so explicitly —
  **that section will need updating when 0.9.4 actually ships.**
- **No 0.9.4 date and no CHANGELOG entry for it.** It is described as "merged to `main`,
  not yet released", phrased to be promotable or deletable in one edit.
- **The 0.9.3 v4-rollup summary is not exhaustive.** #129 merged a large branch; the
  README lists the user-visible items and the CHANGELOG lists a few more, both ending in
  "and several allocator fixes" rather than enumerating every commit.
- **No claim about which fixes were accepted upstream.** Commits reference
  `microsoft/mimalloc#1351` for the bootstrap fix, but that filing postdates the v0.9.0
  tag, so it is not attributed to any release here.
- The 0.8.0 entry's trailing "0.9.0 is reserved for a v3 mimalloc port" line is left in
  place as historical record; the new 0.9.0 entry directly above it resolves it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,38 @@ test suite looks. That risk is real and testing cannot retire it — see
 
 ---
 
+## Release history
+
+Published versions of the [`mimalloc-pprof`](https://crates.io/crates/mimalloc-pprof)
+crate on the v3 line, newest first. Dates are crates.io publish dates (UTC). This table
+is a summary; the full notes, with the reasoning behind each fix, are the record in
+[`rust/mimalloc-pprof/CHANGELOG.md`](rust/mimalloc-pprof/CHANGELOG.md).
+
+| Version | Published | What landed |
+|---|---|---|
+| **0.9.3** | 2026-08-21 | **Fix:** the crate would not build for `aarch64-pc-windows-msvc`. mimalloc's MSVC atomics wrapper reaches for ARM64 `Interlocked` intrinsics that `clang-cl` does not declare, so a `cargo-xwin` cross build failed to compile ([#223](https://github.com/zackees/mimalloc-pprof/issues/223), [#224](https://github.com/zackees/mimalloc-pprof/pull/224)). **Also the first release carrying the v4 line**, merged into `main` after 0.9.2 ([#129](https://github.com/zackees/mimalloc-pprof/pull/129)): upstream engine pin `579f8c0e` -> `bcee5a88` ([#80](https://github.com/zackees/mimalloc-pprof/issues/80), [#113](https://github.com/zackees/mimalloc-pprof/pull/113)), the new experimental `mi_option_purge_zeroes` zero-tracking option ([#67](https://github.com/zackees/mimalloc-pprof/issues/67), [#79](https://github.com/zackees/mimalloc-pprof/pull/79)), the zeroing-`realloc` family exposed from Rust ([#83](https://github.com/zackees/mimalloc-pprof/issues/83), [#108](https://github.com/zackees/mimalloc-pprof/pull/108)), and several allocator fixes. |
+| **0.9.2** | 2026-08-01 | **Fix:** seeded sampling is now actually reproducible. `mi_prof_start_seeded` and `MIMALLOC_PROF_SEED` mixed the ASLR-randomised address of the thread's allocator state into the seed, so a pinned seed produced a different sample sequence in every process while the docs promised determinism ([#91](https://github.com/zackees/mimalloc-pprof/issues/91)). |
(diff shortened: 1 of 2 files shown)


zackees:
Merging. 19 checks green; the single failure is **pre-existing on `main`** and unrelated.

`test (ubuntu-latest)` fails in `benchmark-suite/tests/timed_allocation_audit.rs` with `Rust control-plane allocation entered the measured interval for TinyFixed16`. Main's own `rust-native` run at the same commit (32658015900, `154633e6`) fails the **same** audit — two tests there (`ordinary_measured_region_performs_zero_harness_allocations` and `latency_sampling_buffers_do_not_allocate_after_warmup`) versus one here, which also marks it as nondeterministic rather than a clean signal.

This PR changes two markdown files. It cannot cause an allocation-counting failure.

Worth its own issue — an allocation audit that flaps between 1 and 2 failures across runs is not currently a trustworthy gate.

Issue #277, opened 2026-09-02
ci: cross-compile macOS and Windows test bundles on Linux, execute on one runner per OS (7 macOS + 13 Windows jobs → 1 + 1)


zackees:
_Revised 2026-09-02 after an adversarial review (comments below). Counts, phase table, and toolchains corrected; the reviewer verified all three cross lanes by building this repo on Linux through soldr._

_Revised again 2026-09-02 after **phase B shipped** (PR #282). The section below records
where this body was wrong; the rest of the body is left as written so the corrections are
legible against it._

## Corrections from phase B (measured, not argued)

1. **The Rust Apple builds were already on Linux.** §2/§3 read as if `cross.yml`'s
   `test (x86_64-apple-darwin)` / `test (aarch64-apple-darwin)` also *built* on macOS. They
   do not: `build-cross` has run on `ubuntu-latest` all along and only the `test (*)` rows
   are on Apple runners. "Move the build side to Linux" was therefore a no-op. Artifacts do
   not cross workflows without run-id plumbing, so phase B **duplicates** those two builds
   into `macos-bundles.yml` for the comparison window; the deletion TODO on `build-cross`
   removes the Apple targets there together with the test rows.
2. **The memory-gate baseline key needs `compiler`, not just `arch`.** §3/§4 asked for an
   arch key. Arch alone cannot separate the two lanes: a soldr-cross-built binary and an
   Xcode-built one both report `platform: macos` on the same arm64 `macos-latest` runner.
   `ci/memory_gate.py` takes `--arch` *and* `--compiler`, both optional, and joins them
   into the baseline filename — so `linux-pprof1.json`, `windows-pprof1.json` and
   `macos-pprof1.json` all keep their names and keep matching the runs that produced them.
3. **Today's `ctest-debug-full` compiles Release + `MI_DEBUG=3`, not Debug.** It passes no
   `CMAKE_BUILD_TYPE`, and `CMakeLists.txt` (~137-145) defaults it from the *binary
   directory's name* — `build` does not end in "debug", so it resolves to Release, while
   `MI_DEBUG_FULL` sets `MI_DEBUG=3` independently of the build type. The bundle matches
   that (verified: identical 32-test set either way). A literal `CMAKE_BUILD_TYPE=Debug`
   lane builds locally but is not exercised on CI.
4. **The Apple SDK version must be recorded, never asserted.** soldr's SDK directory on the
[...]

PR #276, opened 2026-09-02
feat: bump upstream overlay pin to 6def7be9 (Bun merge-base) + glibc 2.44 free(NULL) fix

zackees:
Closes #266. Sub-issue of #264 (Bun parity, Phase 1).

## What
- Bump the upstream overlay pin from microsoft/mimalloc `bcee5a88` to `6def7be9` (Bun's merge-base; 205 upstream commits). Selective C-engine overlay per #80: upstream tree taken at `6def7be9`, fork hooks re-applied per file (22 upstream-owned files carried hooks; 10 applied clean, 12 re-derived by hand).
- Carry Bun's `7ac561ab` (real all-NULL empty submap so `free(NULL)` from glibc 2.44's loader works before init) into `src/page-map.c` with provenance, plus `test/test-free-before-init.c` (RED without the fix, GREEN with it; excluded on MSVC, which lacks the constructor attribute).
- `src/static.c` now includes upstream's new `subproc.c` and `prim/prim-tls.c`.
- CLAUDE.md pin record updated; MIMALLOC_FORKS.md rows added.
- Rust: vendored amalgamation regenerated in separate rust-only commits.

## Overlay notes (from the Opus review)
- `theap.c`: the fork's `_mi_theap_free` KNOWN-ISSUE (#78) machinery no longer exists; upstream's `_mi_heap_detach_theaps` implements the try-acquire fix.
- `src/prim/windows/prim.c`: restored to upstream verbatim; upstream fixed the `__GCC__` typo itself (`__MINGW32__` is a real macro), so the fork hunk was a no-op divergence.
- `src/page.c`: upstream `bf054991` enabled `mi_theap_collect_full_pages`; adopted rather than re-disabling on the stale #128 measurement. See #128 comment.
- `alloc.c`/`alloc-aligned.c`: hooks re-threaded through upstream's `_mi_theap_malloc_zero`/ppage rename. `page-map.c`, `arena.c` fork hunks were comment-only.
- The issue's "racy `prev_total == peak` compare" in stats.c does not exist at either pin; nothing removed.
- Inherited semantic change flagged on #266: `_mi_meta_zalloc` now routes through `mi_theap_zalloc`, so allocator-internal metadata fires the alloc hooks. Rule 4 verified intact. Follow-up proposed, not implemented here.
- `test/test-stress.c` taken wholesale; its dropped hunk was dead (`xMI_HEAP_WALK`).

## Local verification
[...]

Issue #171, opened 2026-08-06
perf: publish a bounded four-allocator benchmark dashboard


zackees:
## Context

We want a generated performance dashboard for mimalloc-pprof, modeled on [zackees/zccache's benchmark graphs](https://zackees.github.io/zccache/) and [`benchmark-stats` branch](https://github.com/zackees/zccache/tree/benchmark-stats). The headline comparison is:

- Google TCMalloc
- jemalloc
- pinned upstream mimalloc `dev3`
- mimalloc-pprof

The dashboard must cover common allocation and threading behavior without running the full Cartesian product of allocator × size distribution × lifetime × operation × ownership topology × thread count. A naïve matrix would take hours, obscure the important cases, and risk CI timeouts.

This is a focused publishing/harness deliverable under the broader performance program in #170. The repository currently has a feature-specific zero-tracking A/B benchmark and report, but no general four-allocator suite, historical performance dataset, generated graphs, or Pages dashboard.

Useful prior art already exists:

- zccache publishes generated JPGs, `latest.json`, bounded `history.jsonl`, and `index.html` from an orphan-style `benchmark-stats` branch, embeds the images in its README, and uploads full logs separately as Actions artifacts.
- [`daanx/mimalloc-bench`](https://github.com/daanx/mimalloc-bench) already builds jemalloc, Google TCMalloc (`tcg`), mimalloc, and external allocators; emits `benchres.csv`; and contains relevant allocator workloads such as `alloc-test`, `larson`, `mstress`, `rptest`, `xmalloc-test`, cache-scratch/thrash, large allocation, and thread-lifetime tests.
- #168 establishes native Rust threads as the preferred way to generate new concurrent load. Python may orchestrate builds or render reports, but must not generate the timed allocator workload behind the GIL.

The fork's upstream baseline is the exact overlay pin `dev3@bcee5a88`, not the moving `upstream/dev3` ref. Every competitor must likewise be pinned and recorded so historical results remain interpretable.

### 1. Use named scenario cards instead of a Cartesian product

Define roughly 12–16 versioned scenario cards. Each card intentionally combines one allocation behavior with only the thread topologies that expose distinct allocator mechanisms. Do not independently multiply every parameter.

| Scenario family | Allocation behavior | Thread points/topology | Mechanism covered |
|---|---|---|---|
| Tiny hot path | fixed 16 B and 64 B alloc/free | 1; physical cores | absolute fast-path overhead and local scaling |
[...]

PR #278, opened 2026-09-02
ci: cancel superseded workflow runs per ref (#277 phase 0)

zackees:
Phase 0 of #277. Two things: a `concurrency` group on every workflow, and the tool that
says whether it was worth it.

## Why

Nothing in `.github/workflows/` had a `concurrency` group, so a second push to a PR branch
left the first push's jobs queued and competing for the same runners. Measured over the
last 20 runs of each workflow, **all branches** (full table on #277):

| workflow | runner | n | queue p50 | queue p90 | queue max | exec p50 |
|---|---|---|---|---|---|---|
| `c-unit.yml` | `ubuntu-latest` | 160 | 4s | 19s | 1m10s | 45s |
| `c-unit.yml` | `windows-latest` | 140 | 4s | 24s | 1m14s | 1m46s |
| `c-unit.yml` | `macos-latest` | 60 | 21s | 2m03s | 3m05s | 48s |
| `rust-native.yml` | `macos-latest` | 20 | 27s | 1m32s | 2m07s | 2m31s |
| `cross.yml` | `macos-latest` | 20 | 9m25s | 1h29m | **5h26m** | 8s |
| `cross.yml` | `macos-15-intel` | 20 | 9m23s | 2h32m | **6h13m** | 9s |

The Apple rows on `cross.yml` **execute in 8–14 seconds** and **wait for hours**. macOS is
the motivation; every Linux and Windows row sits at a 4–5 s p50.

This change makes nothing faster. It stops a dead push from holding a macOS slot a live
push needs.

## Per-workflow decisions

**`cancel-in-progress: true`**, group `<stem>-${{ github.ref }}` — pure verification, so a
superseded run's answer is worthless:

| workflow | note |
|---|---|
| `amalgamation-drift.yml` | verification only |
| `asan.yml` | verification only |
| `benchmark-sentinel.yml` | PR-tier sentinel; artifact only, publishes nothing (#171) |
| `c-unit.yml` | the macOS/Windows queue hog |
| `cross.yml` | the 5–6 h Apple rows above |
| `doc-snippets.yml` | verification only |
| `python-lint.yml` | verification only |
| `rust-native.yml` | its only publish-shaped step is `cargo publish --dry-run`, which uploads nothing |
| `zero-tracking.yml` | A/B timing report on a path-filtered PR |
| `fuzz.yml` | **group is `fuzz-${{ github.event_name }}-${{ github.ref }}`** — see below |

**`cancel-in-progress: false`** — cancelling these leaves something half-done:

| workflow | why not cancelled | change |
|---|---|---|
| `auto-release.yml` | `cargo publish` + GitHub Release assets | none (already `false`) |
| `release.yaml` | uploads release assets | **flipped from `true`**; inherited from upstream. A cancelled run leaves the release half-populated. Group also renamed to the file stem for consistency. |
| `stale.yaml` | closes and labels issues and PRs; a half-run leaves the repo partly swept | **added** (`false`) |
| `star-history.yml` | commits a regenerated chart | none (already `false`) |
| `benchmark-stats.yml`, `benchmark-latency.yml`, `benchmark-memory.yml`, `benchmark-scaling.yml` | push the sealed `benchmark-stats` branch and deploy Pages | **none — deliberately left alone.** These four share one group, `benchmark-stats-production`, to serialise *against each other*. Renaming them per-stem would break that. |
| `test.yaml` | inherited upstream workflow, trigger is `push: branches: ['dev*']` — never fires on this fork's branches | none (already `true`, harmless) |

### The one subtle case: `fuzz.yml`

It has a `schedule:` trigger, and a scheduled run carries `github.ref == refs/heads/main`.
With a plain `fuzz-${{ github.ref }}` group, the nightly 2,000,000-iteration run and any
push to `main` would cancel each other — silently turning off the long fuzz run. The group
therefore includes `github.event_name`. Same-event runs on the same ref still supersede.

## `ci/ci_queue_wait.py`

The measurement instrument, so the "after" comparison is the same arithmetic on the same
API fields rather than a vibe:

```
uv run ci/ci_queue_wait.py --limit 20 c-unit.yml cross.yml rust-native.yml
```

Per job: `started_at − created_at` (queue wait) and `completed_at − started_at`
(execution), grouped by the job's `labels` — i.e. what `runs-on:` actually resolved to —
with p50/p90/max (nearest-rank) and the job count per row. Only `gh` is required; no
Python dependencies.

Two things it is deliberate about:

- **The `needs:` caveat is printed, not hidden.** A dependent job is *created* when the run
  is created, so `cross.yml`'s `test (*)` rows measure "pool wait + upstream build".
  `c-unit.yml` and `rust-native.yml` have no `needs:` edges, so their rows are pure pool
  wait. Rows are never pooled across workflows.
- **Cancelled/skipped jobs are excluded** (a cancelled job's `completed_at` is when the
  cancellation landed, not when work finished); **failed jobs are kept**, because a red job
  still occupied a runner for its whole execution.

Fetching is separated from arithmetic, so `ci/tests/test_ci_queue_wait.py` (15 tests)
drives `parse_jobs`/`percentile`/`summarise` from a fixture in the real
`/actions/runs/<id>/jobs` shape, and `--jobs-json` replays that fixture through the CLI.
Nothing in the test suite touches the network.

## Also in this PR

- `docs/ci-gates.md` gains a "Concurrency: superseded runs are cancelled" section with the
  policy, the publisher exclusions, the `fuzz.yml` reasoning, and the measurement command.

## Verification

- `ruff check ci/` and `ruff format --check ci/` clean (ruff 0.12.10, the CI pin).
- `pytest ci/tests` — 96 passed, 15 of them new, 0 failed. (Rebased onto `main` after #276
  merged, which brought `[tool.pytest.ini_options] pythonpath = ["ci"]` — so this PR no
  longer needs to add it — and fixed the stale README-anchor test that was making
  `python-lint` red on `main`.)
- All 19 workflow files re-parse as YAML and every one now has a `concurrency` block.
- **`pyright --strict` is clean.** It initially could not run here (no system Node); `uv run
  --with 'pyright[nodejs]==1.1.411' pyright` ships its own Node and works, which caught four
  real strict errors that are fixed in the last commit. That command is worth knowing for
  anyone else hacking on `ci/` from a Node-less box.

Refs #277. Does not merge anything; phase A (`ci/277-phase-a-test-bundles`) stacks on top.

### DIFF
--- a/.github/workflows/amalgamation-drift.yml
+++ b/.github/workflows/amalgamation-drift.yml
@@ -23,6 +23,11 @@ on:
     branches: [main, v4]
   pull_request:
 
+concurrency:
+  # Pure verification: a superseded push's answer is worthless.
+  group: amalgamation-drift-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check:
     name: vendored C amalgamation matches src/
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -18,6 +18,11 @@ on:
       - 'CMakeLists.txt'
       - '.github/workflows/asan.yml'
 
+concurrency:
+  # Pure verification: a superseded push's answer is worthless.
+  group: asan-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   asan:
     name: address sanitizer
--- a/.github/workflows/benchmark-sentinel.yml
+++ b/.github/workflows/benchmark-sentinel.yml
@@ -17,6 +17,11 @@ on:
 env:
   SOLDR_GITHUB_TOKEN: ${{ github.token }}
 
+concurrency:
+  # PR-tier sentinel; artifact-only, publishes nothing (see #171).
+  group: benchmark-sentinel-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   sentinel:
     name: benchmark-sentinel (${{ matrix.os }})
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -12,6 +12,11 @@ on:
       - 'ci/**'
       - '.github/workflows/c-unit.yml'
 
+concurrency:
+  # The macOS/Windows queue hog (#277). A superseded push's ctest result is worthless.
+  group: c-unit-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   ctest:
     strategy:
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -18,6 +18,11 @@ on:
       - rust/**
       - .github/workflows/cross.yml
 
+concurrency:
+  # Measured ~4 h / ~5.5 h queue waits on the Apple rows (#277).
+  group: cross-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-linux:
     name: build (${{ matrix.target }})
--- a/.github/workflows/doc-snippets.yml
+++ b/.github/workflows/doc-snippets.yml
@@ -23,6 +23,11 @@ on:
       - 'ci/check_doc_snippets.py'
       - '.github/workflows/doc-snippets.yml'
 
+concurrency:
+  # Pure verification: a superseded push's answer is worthless.
+  group: doc-snippets-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check:
     runs-on: ubuntu-latest
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -20,6 +20,13 @@ on:
   schedule:
     - cron: '0 4 * * *'   # nightly, longer budget
 
+concurrency:
+  # event_name is in the group on purpose: a scheduled run carries ref=refs/heads/main,
+  # so without it the nightly 2,000,000-iteration run and a main push would cancel each
+  # other. Same-event runs on the same ref still supersede.
+  group: fuzz-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   fuzz:
     runs-on: ubuntu-latest
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -16,6 +16,11 @@ on:
       - '.github/workflows/python-lint.yml'
       - '.github/workflows/benchmark-*.yml'
 
+concurrency:
+  # Pure verification: a superseded push's answer is worthless.
+  group: python-lint-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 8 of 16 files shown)


zackees:
Heads-up on the red `python-lint`: it is **not** this PR.

`python-lint` runs `pytest ci/tests`, and `test_readme_embeds_only_real_charts_and_clicks_through_to_pages` asserts the README still embeds a chart linking to `.../#throughput`. That embed was removed by a README-only PR, and `python-lint`'s `paths:` filter (`ci/**`, `pyproject.toml`, workflows) means a README-only PR never runs it — so `main` has been red here invisibly. Reproduced on a clean `origin/main` checkout with this branch stashed.

The fix already exists, unmerged, on `bun-parity/p1-pin-bump` (`238218d5`, "update stale README anchor test"). Duplicating it here would just conflict with that branch, so it is deliberately left alone. `ruff check`, `ruff format --check` and `pyright --strict` (0 errors) all pass on this branch.

Worth noting for #277 or a follow-up: a gate whose path filter excludes the files that can break it is the same shape as the seven dead gates in docs/ci-gates.md.

zackees:
Update after rebasing onto `main` (#276 merged mid-flight): `python-lint` is now **green** — that PR carried the stale-README-anchor fix, so the earlier comment above is resolved, and this PR no longer needs its own `pyproject.toml` hunk (#276 added the identical one).

The one remaining red check is `memory-gate (ubuntu-latest)`, and it is the gate calling itself unreliable rather than a regression here — this PR changes no C, no CMake and no test:

```
peak_rss  28.0 MB   baseline 22.3 MB   allowed 25.6 MB
WARNING: observed spread (34.3%) is at or above the tolerance (15%). The threshold
cannot distinguish a regression from noise -- raise RUNS_EXPECTED or the tolerance,
with this measurement as the justification.
FAIL
  - peak_rss regressed: 28.0 MB vs baseline 22.3 MB (+25.6%, allowed +15%)
```

Per docs/ci-gates.md ("a gate that flakes gets ignored, and an ignored gate is worse than none") that warning is the actionable finding, not the FAIL. Deliberately not re-baselined here — the gate says in as many words not to re-baseline a red gate green, and doing so from an unrelated CI PR would be exactly the wrong precedent. Worth its own issue.

zackees:
Opus review: **APPROVE**. Verified group-key uniqueness across all 19 workflows, PR-vs-push isolation, no publisher left on cancel-in-progress, `benchmark-stats-production` serialisation preserved, fuzz keying, and the measurement tool (pyright strict 0, ruff clean, 139 tests) plus arithmetic consistency of the #277 baseline comment (n = 740).

Low follow-ups (not blocking): `ci_queue_wait.py` `--verbose` does not report dropped-job count as its docstring claims; `--limit >100` silently truncates (no pagination); header prints `last N runs` in `--jobs-json` mode; `auto-release.yml`/`test.yaml` still key on `${{ github.workflow }}` rather than the file stem.

Merging with macOS jobs still queued: this PR changes workflows/ci/docs only, Linux and Windows are green, and the only red is the known ubuntu memory-gate drift (deferred per #276).

PR #279, opened 2026-09-02
ci: portable test bundles from ctest json-v1 + serial runner (#277 phase A)

zackees:
Phase A of #277: the portable test-bundle format, the two scripts, and the gate that
proves a bundle runs exactly what ctest runs.

> **Stacked on #278.** Both branches are rebased onto `main` as of #276. Against `main` this
> PR's diff shows #278's two commits first; everything titled "(#277 phase A)" is this PR.
> Merge #278 first and this collapses to the phase-A commits alone.

## What a bundle is

A directory a macOS or Windows runner can execute with **no CMake and no repo checkout**:
every test executable, every `libmimalloc*` / `mimalloc*.dll` / `mimalloc-redirect*.dll`,
and a `tests.json` manifest.

```
uv run ci/bundle_tests.py <build-dir> <out-dir> [--config Debug]
uv run ci/run_test_bundle.py <bundle> [--only NAME…] [--env K=V…]
                                      [--timeout-scale F] [--junit out.xml]
                                      [--compare-junit ctest.xml]
```

One entry per test: `name`, `argv`, `env`, `cwd`, `timeout`, `expect_nonzero`,
`expect_text`, `labels`.

## Lowering rules

`ctest --show-only=json-v1` does not emit a list of plain executables. Both `cmake`
shapes are lowered into manifest fields so the runner needs no CMake:

| ctest command | lowered to |
|---|---|
| `cmake -E env K=V … <exe> [args]` | `env` + the real `argv`; `K=V` tokens consumed up to the first token that is not one |
| `cmake -D TEST_EXE=… -D TEST_ARG=… -D EXPECTED_TEXT=… -P test/run-negative.cmake` | `argv = [TEST_EXE, TEST_ARG?]`, `expect_nonzero: true`, `expect_text: EXPECTED_TEXT`, `timeout: 10` |

The negative-control semantics come straight from `test/run-negative.cmake`, including the
two that are easy to get backwards:

- **A timeout is a failure**, not the expected non-zero exit. The script's own words are
  *"negative control timed out instead of failing fast"*. A hung control proves nothing.
- The expected text is searched in the **combined stdout+stderr**, so a control that fails
  for the *wrong* reason stays red (`"failed for the wrong reason (result=…)"` upstream).

`TEST_ARG` is optional; the effective timeout is the cmake script's 10 s, not the test's
`TIMEOUT 15`.

## Failing loudly

Everything the bundler cannot lower is a hard error **naming the test**, and all problems
are reported in one message rather than one per run:

- an unknown `cmake` argv shape (`cmake -E rm`, an unrecognised `-P` script)
- a `cmake -E env` option (`--unset=`, `--modify`) rather than a plain assignment
- an unrecognised `run-negative.cmake` variable
- a test property outside the allowlist — `ENVIRONMENT`, `TIMEOUT`, `WORKING_DIRECTORY`,
  `LABELS`, `WILL_FAIL`, `DISABLED`. `PASS_REGULAR_EXPRESSION`, `SKIP_RETURN_CODE`,
  `RESOURCE_LOCK` and friends are refused, not silently dropped.
- an executable outside the build tree, a basename collision, or an empty suite

The reason is the one `docs/ci-gates.md` is built around: a bundle that quietly carries
fewer tests than ctest ran reports green while verifying less. That is precisely the shape
of all seven dead gates this repo has already found.

## Path handling

Absolute build-tree paths appear in **argv and in env values** (`LD_PRELOAD=/abs/build/…`,
`DYLD_INSERT_LIBRARIES=/abs/build/…`); both are rewritten to a `${BUNDLE}` placeholder the
runner expands. Files are **flattened** into the bundle root, because a Windows DLL must
sit beside the exe that loads it and a flat directory is what lets
`LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH` override the build-tree RPATH baked into the
executables. Symlinks are recreated as symlinks so the `libmimalloc-debug.so →
.so.3 → .so.3.4` SONAME chain survives.

## The `bundle-roundtrip` job (c-unit.yml, ubuntu)

`MI_PPROF=ON MI_DEBUG_FULL=ON`, mirroring the existing `ctest-debug-full` ubuntu row:
configure → build → `ctest --output-junit ctest.xml` (the reference) → bundle → **`mv build
build.moved-away`** → replay with `--compare-junit`. The bundle and both JUnit files upload
as an artifact.

The `mv` is the load-bearing step. The executables carry an RPATH into the build tree, so
without it a broken bundle passes on the build machine by loading the original libraries
and the portability claim goes untested on the only machine that can check it.

`--compare-junit` compares against `ctest --output-junit` rather than parsing
`LastTest.log`: it is a stable documented format, so the comparison is deterministic and
testable from fixtures. It asserts the same name set **and** the same pass/fail per test.

## Local verification (this machine, Linux x86-64)

Both runs had the build tree renamed away before replaying:

```
# Debug, MI_PPROF=ON MI_DEBUG_FULL=ON
bundled 31 tests and 26 files into out/bundle-a-out
  6 negative control(s) lowered from run-negative.cmake
  5 test(s) carry environment overrides
100% tests passed, 0 tests failed out of 31
Total Test time (real) =    19.86 sec
bundle matches out/ctest-debug.xml: same 31 test names, same results

# Release, MI_PPROF=ON
bundled 24 tests and 24 files into out/bundle-rel-out
  5 test(s) carry environment overrides
100% tests passed, 0 tests failed out of 24
Total Test time (real) =    26.63 sec
bundle matches out/ctest-rel.xml: same 24 test names, same results
```

## Corrections to #277

- §1 says *"verified on the `MI_DEBUG_FULL` tree, 12 of 32 tests invoke `cmake`: 6 are
  `cmake -E env` and 6 are `cmake -P run-negative.cmake`"*. **Actual: 31 tests, 7 of which
  invoke `cmake` — 6 negative controls plus exactly one `cmake -E env`
  (`test-stress-dynamic`).** The Release tree has 24 tests and none. The lowering rules the
  issue asks for are all still needed; only the counts were off.
- The rest of §1 held up: no `WILL_FAIL` and no `PASS_REGULAR_EXPRESSION` in use; the live
  properties are `ENVIRONMENT`, `TIMEOUT`, `WORKING_DIRECTORY`; absolute paths really do
  appear in env values. `WILL_FAIL` and `DISABLED` are implemented anyway, since they are
  one line each and the alternative is a future `add_test` breaking the bundler.

## Tests and checks

- `ci/tests/test_test_bundles.py`, 42 tests: the two lowering rules, every refusal path,
  path relativisation in argv and in env (`LD_PRELOAD` and `DYLD_INSERT_LIBRARIES`),
  basename collisions, the negative-control semantics end to end (wrong reason → red, zero
  exit → red, timeout → red *even with* `expect_nonzero`), loader-path injection,
  `--only`/`--env`/`--timeout-scale`/`--junit`/`--compare-junit`, and a roundtrip on a tiny
  fake bundle whose build tree is **deleted** before replay.
- `pytest ci/tests`: 139 passed, 0 failed.
- `ruff check ci/` + `ruff format --check ci/` clean on the CI pin (0.12.10).
- **`pyright --strict` is clean.** It initially could not run here (no system Node); `uv run
  --with 'pyright[nodejs]==1.1.411' pyright` ships its own Node and works, which caught four
  real strict errors that are fixed in the last commit. That command is worth knowing for
  anyone else hacking on `ci/` from a Node-less box.
- `uv run ci/verify_local.py --list` — `ci/verify_local.py` did not exist on `main` when
  this work started; #276 landed it mid-flight, along with a drift gate that fails when a
  Linux CI job references a `ci/*.py` the local mirror does not. So `bundle-roundtrip` now
  has a matching `bundle` config, and the gate is reproducible locally:

```
$ uv run ci/verify_local.py --only bundle
[  103.6s] bundle       PASS
…
100% tests passed, 0 tests failed out of 32          # ctest reference
bundled 32 tests and 27 files into …/verify/bundle/bundle
100% tests passed, 0 tests failed out of 32          # the bundle, build tree removed
bundle matches …/verify/bundle/ctest.xml: same 32 test names, same results
```
- CI uses `python3`, matching every existing workflow here; the scripts are stdlib-only, so
  `uv run` behaves identically.

## One CI-only bug this caught

The first CI run failed and it was worth having: `ctest --output-junit` resolves a
**relative** path against the *build* directory, not the working directory, so the
reference `ctest.xml` landed in `build/` and the `mv` carried it away. Locally the roundtrip
had always been run with an absolute path, so it never showed. The workflow now passes
`$GITHUB_WORKSPACE/ctest.xml`, `run_test_bundle.py` names the gotcha instead of leaking an
ElementTree stack trace, `ctest_run()` in `verify_local.py` resolves the path so no caller
can repeat it, and there is a test. The bundle itself ran **31/31 green with the build tree
gone** on that failing run — the thing the job exists to prove.

## Not in scope

No toolchain work: the cross lanes are phase B/C/D (soldr lanes, #277 §2). Nothing outside
`ci/`, `.github/workflows/c-unit.yml` and `docs/ci-gates.md` is touched, so rule 2 holds.

Refs #277.

### DIFF
--- a/.github/workflows/amalgamation-drift.yml
+++ b/.github/workflows/amalgamation-drift.yml
@@ -23,6 +23,11 @@ on:
     branches: [main, v4]
   pull_request:
 
+concurrency:
+  # Pure verification: a superseded push's answer is worthless.
+  group: amalgamation-drift-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check:
     name: vendored C amalgamation matches src/
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -18,6 +18,11 @@ on:
       - 'CMakeLists.txt'
       - '.github/workflows/asan.yml'
 
+concurrency:
+  # Pure verification: a superseded push's answer is worthless.
+  group: asan-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   asan:
     name: address sanitizer
--- a/.github/workflows/benchmark-sentinel.yml
+++ b/.github/workflows/benchmark-sentinel.yml
@@ -17,6 +17,11 @@ on:
 env:
   SOLDR_GITHUB_TOKEN: ${{ github.token }}
 
+concurrency:
+  # PR-tier sentinel; artifact-only, publishes nothing (see #171).
+  group: benchmark-sentinel-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   sentinel:
     name: benchmark-sentinel (${{ matrix.os }})
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -12,6 +12,11 @@ on:
       - 'ci/**'
       - '.github/workflows/c-unit.yml'
 
+concurrency:
+  # The macOS/Windows queue hog (#277). A superseded push's ctest result is worthless.
+  group: c-unit-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   ctest:
     strategy:
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -133,7 +133,10 @@ jobs:
       - run: cmake -B build -DMI_PPROF=ON -DMI_DEBUG_FULL=ON
       - run: cmake --build build --config Debug
       - name: ctest (the reference the bundle must reproduce)
-        run: ctest --test-dir build -C Debug --output-on-failure --output-junit ctest.xml
+        # --output-junit resolves relative to the *build* directory, not the working
+        # directory, so an absolute path is required here: a relative one would land in
+        # build/ and be carried off by the `mv` below.
+        run: ctest --test-dir build -C Debug --output-on-failure --output-junit "$GITHUB_WORKSPACE/ctest.xml"
       - name: Build the bundle
         run: python3 ci/bundle_tests.py build bundle --config Debug
       - name: Move the build tree out of the way
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -117,6 +117,41 @@ jobs:
       - uses: actions/checkout@v4
       - run: cmake -B build -DMI_PPROF=ON -DMI_DEBUG_FULL=ON && cmake --build build --config Debug && ctest --test-dir build -C Debug --output-on-failure
 
+  bundle-roundtrip:
+    # Issue #277 phase A. The plan in #277 is to build the macOS/Windows test binaries on
+    # Linux and ship them to one runner per OS as a portable *test bundle*. That is only
+    # credible if the bundle runs exactly the tests ctest runs and gets exactly the same
+    # results, so this job proves the equivalence on the config where every awkward
+    # command shape lives: MI_DEBUG_FULL emits 6 `cmake -P run-negative.cmake` negative
+    # controls and a `cmake -E env` LD_PRELOAD wrapper, none of which is a plain argv.
+    #
+    # The reference run and the bundle run are compared by name AND by pass/fail, so a
+    # bundle that silently drops a test fails here rather than reporting green on less.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: cmake -B build -DMI_PPROF=ON -DMI_DEBUG_FULL=ON
+      - run: cmake --build build --config Debug
+      - name: ctest (the reference the bundle must reproduce)
+        run: ctest --test-dir build -C Debug --output-on-failure --output-junit ctest.xml
+      - name: Build the bundle
+        run: python3 ci/bundle_tests.py build bundle --config Debug
+      - name: Move the build tree out of the way
+        # The test executables carry an RPATH into the build tree. Without this the bundle
+        # would "pass" by loading the original libraries off disk, and the whole
+        # portability claim would be untested on the only machine that can check it.
+        run: mv build build.moved-away
+      - name: Replay the bundle and require identical results
+        run: python3 ci/run_test_bundle.py bundle --compare-junit ctest.xml --junit bundle.xml
+      - uses: actions/upload-artifact@v4
+        with:
+          name: test-bundle-ubuntu-debug-full
+          path: |
+            bundle
+            ctest.xml
+            bundle.xml
+          retention-days: 7
+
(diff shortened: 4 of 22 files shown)


zackees:
Heads-up on the red `python-lint`: it is **not** this PR.

`python-lint` runs `pytest ci/tests`, and `test_readme_embeds_only_real_charts_and_clicks_through_to_pages` asserts the README still embeds a chart linking to `.../#throughput`. That embed was removed by a README-only PR, and `python-lint`'s `paths:` filter (`ci/**`, `pyproject.toml`, workflows) means a README-only PR never runs it — so `main` has been red here invisibly. Reproduced on a clean `origin/main` checkout with this branch stashed.

The fix already exists, unmerged, on `bun-parity/p1-pin-bump` (`238218d5`, "update stale README anchor test"). Duplicating it here would just conflict with that branch, so it is deliberately left alone. `ruff check`, `ruff format --check` and `pyright --strict` (0 errors) all pass on this branch.

Worth noting for #277 or a follow-up: a gate whose path filter excludes the files that can break it is the same shape as the seven dead gates in docs/ci-gates.md.

zackees:
Opus review: **APPROVE**. Roundtrip re-run by the reviewer: Debug FULL 32/32 and Release 25/25, bundle matches ctest by name and result both ways; zero absolute paths in the produced `tests.json`; lowering semantics match `run-negative.cmake` and `cmake -E env` (values with `=` preserved); unrecognised argv shapes refuse loudly; junit relative-path fix correct; pyright strict/ruff/pytest clean.

Follow-ups filed on #277 (Phase A polish, before Phase B ships bundles cross-host):
- **Medium:** only `argv[0]` is checked for the bundle placeholder; an absolute source-tree path in `argv[1:]`, an env value, or cwd passes through verbatim. Add a final scan of argv/env/cwd for any `os.path.isabs` string not under the placeholder → `problems`.
- Low: `_lower_run_negative` uses `if arg:` (empty `TEST_ARG` dropped; use `"TEST_ARG" in defines`); negative timeout should be `min(10, TIMEOUT)`; `copy_into` should assert symlink targets are bare basenames; `os.pathsep`-joined build paths in env die with `FileNotFoundError` instead of `BundleError`; `generated_from.build_dir` is the one absolute host path left in the manifest; docs show the spaced `-D TEST_EXE=` form the bundler refuses; test counts hard-coded in docs; no tests for spaced `-D` refusal or empty `TEST_ARG`; Windows drive-letter case in `_normalise`.

Merging with macOS jobs still queued for the same reason as #278.

Issue #266, opened 2026-09-01
Bun parity P1: bump upstream overlay pin bcee5a88 -> 6def7be9 (Bun's merge-base) + carry glibc 2.44 free(NULL) fix


zackees:
**Depends on:** nothing (#265 can land in parallel). **Blocks:** #267, #270, #271, #272.

Sub-issue of #264 (Phase 1). **Prerequisite for Phases 2, 5, 6, 7.** Follows the process recorded in #66 / #80 (last bump, `579f8c0e` → `bcee5a88`).

### Why
Our overlay is pinned to `upstream/dev3` `bcee5a88`. Bun's fork is based on `6def7be9` (2026-08-09), **205 upstream commits later** (`git log --oneline bcee5a88..6def7be9 | wc -l` = 205). Everything Bun added that we need — scavenger, hole purging, atfork handlers, the heap-teardown protocol — is written against two things that do not exist at our pin: the 2-level page map (`d63979ae`) and `src/subproc.c` (verified: `git ls-tree 6def7be9 src/subproc.c` exists, `git ls-tree bcee5a88 src/subproc.c` is empty). Nothing downstream can be ported first.

### Target
Bump to **`6def7be9`** exactly (Bun's merge-base), not the live `dev3` tip. Matching Bun's base byte-for-byte is what makes their diffs apply in later phases. A later bump to tip is a separate issue.

### Steps
1. `git fetch upstream dev3`. Read `git log --oneline --stat bcee5a88..6def7be9 -- src/alloc.c src/free.c src/page.c src/page-map.c src/theap.c src/heap.c src/init.c src/stats.c include/mimalloc/types.h include/mimalloc/internal.h` first; these are where our hooks live.
[...]

PR #282, opened 2026-09-02
ci: build macOS test bundles on Linux via soldr, run them on one macOS runner (#277 phase B)

zackees:
Closes the macOS half of #277. Six macOS jobs per push become one.

The motivation is queue wait, not build time: over 20 runs per workflow (#277 phase 0),
`cross.yml`'s Apple rows **executed in 8-14 s and waited up to 5h26m / 6h13m**, p90
1h29m / 2h32m, while every ubuntu row waited 4-5 s.

## What lands

| new job | runner | replaces |
|---|---|---|
| `build-macos (macos-arm64-release)` | ubuntu-latest | — (produces the bundle) |
| `build-macos (macos-arm64-debug-full)` | ubuntu-latest | — |
| `build-macos (macos-arm64-leak)` | ubuntu-latest | — (memory-gate positive control) |
| `build-rust (aarch64-apple-darwin)` | ubuntu-latest | — |
| `build-rust (x86_64-apple-darwin)` | ubuntu-latest | — |
| `run-macos` | **macos-latest (one VM)** | `ctest (macos-latest)`, `ctest-debug-full (macos-latest)`, `memory-gate (macos-latest)`, `cross` `test (aarch64-apple-darwin)`, `cross` `test (x86_64-apple-darwin)`, `rust-native` `test (macos-latest)` |

Those six are now **`continue-on-error: true`** — the control arm, for at least ten
pushes — each with a dated TODO naming what deletes it.

## Toolchain spike (run locally on Linux, soldr 0.9.11, clang 21.1.5)

`cmake/toolchains/soldr-aarch64-apple-darwin.cmake` consumes only soldr's exported env
(`CC_`/`CXX_`/`CFLAGS_`/`CXXFLAGS_`/`AR_`/`RANLIB_` per triple, `SDKROOT`). Both configs
built clean:

```
-- Architecture: arm64
-- Architecture specific optimization is enabled (with -march=armv8.1-a) (MI_OPT_ARCH=ON)
-- Compiler defines : ...;MI_OSX_ZONE=1;MI_OSX_INTERPOSE=1;MI_PPROF=1;...
-- Link libraries   : pthread
```

`-march=armv8.1-a` is accepted by soldr's clang 21 (the #277 review's zig lane rejected
it). `Link libraries : pthread` is the `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY` fix
working: without it `find_link_library()` falls through to `find_library()` and hands the
host's ELF `librt.so`/`libatomic.so` to a Mach-O link.

`llvm-objdump --macho` on the produced dylib (no `llvm-otool` exists in soldr's toolchain):

```
MH_MAGIC_64   ARM64  ALL  0x00  DYLIB  ... MH_HAS_TLV_DESCRIPTORS
  cmd LC_BUILD_VERSION       platform macos   sdk 15.5   minos 11.0
  cmd LC_ID_DYLIB            name @rpath/libmimalloc.3.dylib
  cmd LC_CODE_SIGNATURE      dataoff 267808  datasize 2256
sections:  __DATA_CONST,__mod_init_func   __DATA,__interpose   __DATA,__thread_vars
```

`LC_CODE_SIGNATURE` (ad-hoc, from `ld64.lld`) matters: arm64 macOS kills an unsigned image
before `main()` with no diagnostic. All test executables link only `/usr/lib/libSystem.B.dylib`
(static mimalloc); `test-stress-dynamic` is the one that loads the dylib, via
`DYLD_INSERT_LIBRARIES`.

Bundling both trees with `ci/bundle_tests.py`:

```
bundled 25 tests and 25 files      (release)
bundled 32 tests and 27 files      (debug-full)
  6 negative control(s) lowered from run-negative.cmake
  10 test(s) carry environment overrides
"DYLD_INSERT_LIBRARIES": "${BUNDLE}/libmimalloc.3.5.dylib"
```

Zero absolute paths in either `tests.json`; `.dylib` entries present with the
`libmimalloc.dylib -> libmimalloc.3.dylib` symlink chain intact through a tar round-trip
(exec bits and symlinks both survive — which is why the bundle is tarred: `upload-artifact@v4`
preserves neither).

**I cannot execute Mach-O on this host.** `run-macos` on this PR is the proof.

## Coverage

`ci/bundle_coverage.py` runs on every `run-macos` and fails if a test the native `ctest`
would run is missing. Run locally against the cross trees (CI compares against
**Xcode-clang-configured** trees, which is the part only CI can do):

| config | native | bundle | missing from bundle | extra in bundle |
|---|---:|---:|---|---|
| ctest (macos-latest) | 25 | 25 | — | — |
| ctest-debug-full (macos-latest) | 32 | 32 | — | — |

Stated, not gated, because nothing can gate them:

- the **8 doctests** in `rust/mimalloc-pprof/src/lib.rs` run under `cargo test` and cannot
  run from a `--tests` binary — Linux-only after this (#277 §4)
- the native compile-compat step uses Xcode clang and runs **no tests**: it catches
  compile-only breakage, not codegen/runtime differences
- `cross.yml`'s Apple rows have `pprof: false`, so `run-macos` has no
  pprof-parses-a-dump step either — parity, not a reduction

## Memory gate

`memory_gate.py` gained `--arch`/`--compiler`, which join the baseline file's **name**.
Both optional, so `linux-pprof1.json`, `windows-pprof1.json` and `macos-pprof1.json` keep
their names and keep matching the runs that produced them — **the ubuntu baseline is
untouched**. `macos-pprof1.json` is annotated `arch: arm64`, `compiler: apple-clang` so a
reader can tell which toolchain recorded 13.4 MB.

The soldr lane asks for `macos-arm64-soldr-clang-21-pprof1.json`, which does not exist, so
its **first run is deliberately yellow**: `::warning:: No committed baseline` and exit 0,
with the per-run JSON uploaded as `memory-gate-macos-arm64-soldr`. Its positive control is
skipped with a warning until that baseline is committed (`control` requires `check` to
*fail*, and a missing baseline is not a failure). Baselining it is a deliberate follow-up
with the numbers, not something this PR does silently.

## Also in here

- **Phase A review follow-up (medium):** `bundle_tests.py` validated only `argv[0]`.
  Every lowered test is now scanned across argv, env values (split on `os.pathsep`) and
  cwd; Windows drive/UNC paths count as absolute regardless of host, because every cross
  bundle is built on Linux where `os.path.isabs("C:/build/x")` is False.
  `generated_from.build_dir` is reduced to a basename — it was the one host path left in
  the manifest.
- **#277 §4's pre-existing gap:** `build-cross` had none of the frame-pointer flags
  `build-linux` sets, while the profiler walks frame pointers off Windows. Fixed in both
  `cross.yml` and the new workflow — by *appending* to `CFLAGS_<triple>` and
  `CARGO_ENCODED_RUSTFLAGS`, since `soldr prepare` exports both and they shadow the plain
  names, with an assertion step because a flag that never reaches the compiler is this
  repo's most-repeated CI bug.

## Local verification

`uv run ci/verify_local.py --keep-going`:

```
release     PASS    off        PASS    debug-full PASS    guarded  PASS
shared      PASS    bundle     PASS    memory-gate PASS   diag     PASS
rust        PASS    lint       FAIL*   asan       FAIL*
```

\* both environmental on this host, not from this branch: `asan` has no
`sanitizer/asan_interface.h` installed, and `lint` runs `uv run --with pyright==1.1.411`,
whose `nodeenv` download fails with an SSL cert error here. Run with the packaged node
(`--with 'pyright[nodejs]==1.1.411'`) pyright reports **0 errors**; `ruff check`,
`ruff format --check` and `pytest ci/tests` (167 passed) are clean.

## Things #277's body got wrong (load-bearing for phases C/D)

1. **The Rust Apple builds were already on Linux** (`cross.yml` `build-cross`), so "move
   the build side to Linux" was a no-op. Artifacts do not cross workflows without run-id
   plumbing, so they are duplicated into `macos-bundles.yml` for the comparison window and
   the TODO on `build-cross` says to drop the Apple targets with the test rows.
2. **The memory-gate key needs `compiler`, not just `arch`.** Native and soldr binaries
   both report `platform: macos` on the same arm64 runner; arch alone cannot separate them.
3. **Native `ctest-debug-full` compiles Release + `MI_DEBUG=3`**, not Debug: it passes no
   `CMAKE_BUILD_TYPE` and CMakeLists.txt defaults it from the *binary directory name*. The
   bundle matches that (verified: identical 32-test set either way).
4. **soldr's SDK directory is 14.5 but `LC_BUILD_VERSION` records sdk 15.5** (the review
   saw 15.5). The job records it into the step summary and asserts nothing against a constant.
5. **No `llvm-otool`** in soldr's toolchain; `llvm-objdump --macho` covers everything needed.
6. `NotRequired` is unavailable under this repo's pyright `pythonVersion = "3.9"`, so the
   two optional result keys are read through a narrowing accessor instead.

---

## Review round 2 (cfcec424)

**Medium fixed.** `run-macos`'s gate steps read `memory_gate.py check`'s exit 2 as "no
baseline yet, warn and pass", but `check` exits 2 for any caught `OSError`/`ValueError`/
`KeyError` as well, and the eight runs sat under `set +e` — so a crashed
`mimalloc-test-memory-gate` would leave the glob unexpanded and produce a **green** step
with a reassuring `::warning::No committed baseline`. Same shape on the leak-control path,
which also uploaded nothing. Now: `where` exits **3** for "no baseline" and keeps **2** for
"cannot read this run", both steps branch on that probe (any other rc is a hard failure),
both require all eight JSON files before calling the gate at all (counted with a glob —
under `set -o pipefail`, `ls … | wc -l` fails when it finds nothing and would kill the step
before it could say why), and `leak-*.json` is uploaded alongside `result-*.json`.

Lows in the same commit: the toolchain's `FATAL_ERROR` expanded the missing variable's
empty value instead of naming it (now `SDKROOT is not set: run soldr prepare …`, verified
by configuring with the variables unset); `run-macos` gains `timeout-minutes: 30`; the two
native coverage configures now name the `c-unit.yml` job each mirrors, including why
`Release` is not a typo for the debug-full one.

**Dead retry removed from the new job only.** `${cmd/soldr/soldr --no-cache}` substitutes
nothing when `cmd` is `cargo test …` with no `soldr` in it, so the soldr#1969 "retry" re-ran
the identical command and reported the same failure twice. Removed in
`macos-bundles.yml`; **`cross.yml` carries the same dead retry and is deliberately left
alone**, so this PR does not change the control arm of the comparison window. Worth a
follow-up issue.

**Untested configuration, stated:** `-DCMAKE_BUILD_TYPE=Debug` with this toolchain builds
and bundles cleanly *locally* (32 tests, identical name set to Release+`MI_DEBUG_FULL`),
but no CI job configures it — all three bundles are `Release` because that is what the
native jobs compile. A literal Debug lane is unexercised on CI.

**Deferred to phase C, recorded here so it is not lost:** `ci/bundle_tests.py`'s
leaked-absolute-path scan matches a bare absolute path, so it misses flag-prefixed ones
(`--out=/abs`, `-I/abs`) and a `C:\…` element inside a value that Linux's `os.pathsep`
(`:`) splits apart. Neither shape occurs in today's manifests; both become reachable when
phase C starts producing Windows bundles on Linux.

### DIFF
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -24,6 +24,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    # TODO(#277 phase B, added 2026-09-02): DELETE the macos-latest row after ten pushes
+    # have compared it against `run-macos` in macos-bundles.yml. Until then it is the
+    # control arm and is informational only -- a permanently-yellow job is worse than an
+    # absent one, so this comment carries its own expiry.
+    continue-on-error: ${{ matrix.os == 'macos-latest' }}
     steps:
       - uses: actions/checkout@v4
       - run: cmake -B build -DMI_PPROF=ON
@@ -113,6 +118,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    # TODO(#277 phase B, added 2026-09-02): DELETE the macos-latest row after ten pushes
+    # have compared it against `run-macos` in macos-bundles.yml. Until then it is the
+    # control arm and is informational only -- a permanently-yellow job is worse than an
+    # absent one, so this comment carries its own expiry.
+    continue-on-error: ${{ matrix.os == 'macos-latest' }}
     steps:
       - uses: actions/checkout@v4
       - run: cmake -B build -DMI_PPROF=ON -DMI_DEBUG_FULL=ON && cmake --build build --config Debug && ctest --test-dir build -C Debug --output-on-failure
@@ -257,6 +267,15 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    # TODO(#277 phase B, added 2026-09-02): DELETE the macos-latest row after ten pushes
+    # have compared it against `run-macos` in macos-bundles.yml. Until then it is the
+    # control arm and is informational only -- a permanently-yellow job is worse than an
+    # absent one, so this comment carries its own expiry.
+    # The macOS row keeps matching ci/memory-baselines/macos-pprof1.json, which is now
+    # annotated `compiler: apple-clang`: it declares no --arch/--compiler, so its baseline
+    # filename is unchanged. The soldr lane asks for its own file instead of borrowing
+    # this number (#277 phase B).
+    continue-on-error: ${{ matrix.os == 'macos-latest' }}
     steps:
       - uses: actions/checkout@v4
       - run: cmake -B build -DMI_PPROF=ON
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -133,6 +133,47 @@ jobs:
         env:
           TARGET: ${{ matrix.target }}
         run: soldr prepare --target "$TARGET" --github-env "$GITHUB_ENV"
+      - name: Keep the Apple builds unwindable (frame pointers)
+        # #277 §4: build-linux sets RUSTFLAGS/CFLAGS at job level and this job set
+        # neither, so the Apple test binaries were built without frame pointers while the
+        # profiler walks frame pointers on every non-Windows platform. Copying
+        # build-linux's job-level `env:` here would be dead -- `soldr prepare` has already
+        # exported CARGO_ENCODED_RUSTFLAGS (which cargo lets shadow RUSTFLAGS entirely)
+        # and CFLAGS_<triple> (which cc-rs prefers over plain CFLAGS) -- so both are
+        # appended to. Apple targets only: the profiler does not walk frame pointers on
+        # Windows, and whether clang-cl accepts the flag is a #277 phase D question.
+        #
+        # Kept in step with macos-bundles.yml's build-rust-apple: while the Apple rows
+        # below are the informational control arm for #277 phase B, both arms must build
+        # the same way or a divergence has two possible causes instead of one.
+        if: contains(matrix.target, 'apple-darwin')
+        env:
+          TARGET: ${{ matrix.target }}
+        run: |
+          set -euo pipefail
+          under="${TARGET//-/_}"
+          eval "existing=\${CFLAGS_${under}:-}"
+          echo "CFLAGS_${under}=${existing} -fno-omit-frame-pointer" >> "$GITHUB_ENV"
+          if [ -n "${CARGO_ENCODED_RUSTFLAGS:-}" ]; then
+            printf 'CARGO_ENCODED_RUSTFLAGS=%s\037-Cforce-frame-pointers=yes\n' \
+              "$CARGO_ENCODED_RUSTFLAGS" >> "$GITHUB_ENV"
+          else
+            echo 'CARGO_ENCODED_RUSTFLAGS=-Cforce-frame-pointers=yes' >> "$GITHUB_ENV"
+          fi
+      - name: The frame-pointer flags must actually be in the build's environment
+        # A flag that never reaches the compiler is this repository's most-repeated CI bug
+        # (MI_GUARDED, docs/ci-gates.md). Assert on the environment the next step runs
+        # with, not on what the previous step intended to write.
+        if: contains(matrix.target, 'apple-darwin')
+        env:
+          TARGET: ${{ matrix.target }}
+        run: |
+          set -euo pipefail
+          eval "cflags=\${CFLAGS_${TARGET//-/_}:-}"
+          case "$cflags" in *-fno-omit-frame-pointer*) ;; *)
+            echo "::error::CFLAGS_${TARGET//-/_} lost -fno-omit-frame-pointer: '$cflags'"; exit 1;; esac
+          case "${CARGO_ENCODED_RUSTFLAGS:-}" in *force-frame-pointers*) ;; *)
+            echo "::error::CARGO_ENCODED_RUSTFLAGS lost -Cforce-frame-pointers"; exit 1;; esac
       - name: Build test binaries through Soldr
         env:
           TARGET: ${{ matrix.target }}
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -236,6 +236,12 @@ jobs:
     name: test (${{ matrix.target }})
     needs: [build-linux, build-cross, build-win-gnu]
     runs-on: ${{ matrix.runner }}
+    # TODO(#277 phase B, added 2026-09-02): DELETE the two apple-darwin rows -- and the
+    # matching targets from build-cross -- after ten pushes have compared them against
+    # macos-bundles.yml's `run-macos`, which runs the same binaries on one runner instead
+    # of two. These are the rows that measured 8-14 s of execution behind up to 5h26m /
+    # 6h13m of queue wait. Informational until then.
+    continue-on-error: ${{ contains(matrix.target, 'apple-darwin') }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/macos-bundles.yml
+++ b/.github/workflows/macos-bundles.yml
@@ -1,0 +1,411 @@
+name: macos-bundles
+
+# Issue #277 phase B: build the macOS test binaries on Linux, execute them on ONE macOS
+# runner.
+#
+# The motivation is not build time. Measured over 20 runs per workflow (#277 phase 0),
+# cross.yml's Apple rows EXECUTE in 8-14 seconds and WAIT up to 5h26m (arm64) / 6h13m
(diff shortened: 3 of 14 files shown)


zackees:
## CI results (run 33598627858)

`run-macos` **passed in 1m40s** on one `macos-latest` VM, from binaries built entirely on Linux:

```
100% tests passed, 0 tests failed out of 25    (macos-arm64-release,    27.03 s)
100% tests passed, 0 tests failed out of 32    (macos-arm64-debug-full, 24.40 s)
```

Both Apple Rust test-bin sets ran (x86_64 under Rosetta — `arch -x86_64 /usr/bin/true`
already worked, so no install was needed on this image).

`build-macos` proved the artifact before shipping it, on the runner, not just locally:

```
-- Link libraries   : pthread
MH_MAGIC_64  ARM64  ALL  0x00  DYLIB  ...  MH_HAS_TLV_DESCRIPTORS
       cmd LC_BUILD_VERSION   platform macos   sdk 15.5
bundled 25 tests and 25 files into bundle
```

(soldr's SDK directory on the runner is `apple-sdk/14.5/...`; `LC_BUILD_VERSION` records
sdk 15.5. Recorded, never asserted against a constant.)

Coverage, against trees configured on the runner by **Apple clang 21.0.0
(Xcode_26.6, `arm64-apple-darwin25.5.0`)** — the bundles were built by **soldr's clang 21.1.5**:

| config | native | bundle | missing from bundle | extra in bundle |
|---|---:|---:|---|---|
| ctest (macos-latest) | 25 | 25 | — | — |
| ctest-debug-full (macos-latest) | 32 | 32 | — | — |

Memory gate took the intended bootstrap path:

```
No baseline at .../ci/memory-baselines/macos-arm64-soldr-clang-21-pprof1.json.
  runs (MB)              13.3  13.3  13.3  13.3  13.3  13.3  13.3  13.3
  spread across runs     0.0%
::warning::No committed baseline for the soldr arm64 lane yet
```

**13.3 MB min-of-8, 0.0% spread**, against Apple clang's committed 13.4 MB min-of-4 on the
same runner — i.e. the cross-built binary is indistinguishable from the native one on the
gated metric. Committing that baseline is deliberately left as a follow-up rather than
done silently inside this PR.

## One red check, pre-existing on `main`

`memory-gate (ubuntu-latest)` failed: 27.3 MB min-of-8 vs baseline 22.3 MB, **spread 22.0%**,
with the script's own warning that "the threshold cannot distinguish a regression from
noise". This branch contains **zero C changes** (only `ci/`, `.github/`, `cmake/toolchains/`,
`docs/`, `CLAUDE.md`), so it cannot have moved peak RSS, and the log confirms the ubuntu
lane still matches its original file (`baseline=linux+pprof1 ... (linux-pprof1.json)`) —
the arch/compiler keying did not move it.

`main`'s own last five completed runs of this job: **failure, failure, failure, success,
success** (runs 33591755528, 33565889274, 33546799862 red; 33594374977, 33546524584 green).
That is exactly the flake `ci/memory_gate.py`'s own comment block documents — five
PASS/FAIL flips on identical Release object code — and it is why `RUNS_EXPECTED` went 4 → 8
in #266. min-of-8 has not fixed it. Reported rather than papered over (CLAUDE.md rule 7);
re-baselining or raising the tolerance is a separate, measured decision and not something
a macOS-CI PR should make.

zackees:
Review round 2 pushed: **cfcec424** (on top of bfdceff3). All lows in the same commit.

Medium: `where` now exits **3** = no baseline, **2** = cannot read this run; both gate steps branch on that probe (any other rc is a hard error), both require all 8 JSON files first (glob-counted, not `ls | wc -l`, which fails under `pipefail` when empty), and `leak-*.json` is uploaded. Verified on run 33600566787: run-macos green 1m57s, 25/25 + 32/32, rc=3 path taken, both artifacts uploaded.

Lows: FATAL_ERROR names the variable (`SDKROOT is not set: …`); `timeout-minutes: 30`; coverage configures cross-referenced to their `c-unit.yml` jobs; dead `${cmd/soldr/…}` retry removed here only — `cross.yml`'s copy left alone. Deferred scan gaps + the untested Debug lane are recorded in the PR body. Issue #277 body updated with all eight corrections.

One red: `test (windows-latest)` (rust-native) — `timed_allocation_audit.rs:195`, "Rust control-plane allocation entered the measured interval for TinyFixed16", left 1 right 0. This branch touches **zero** `rust/` files, it passed on the previous push of this same branch, and `main`'s last five runs of that job are failure/success/cancelled/failure/success (33600219845, 33594374922, 33591755521 …). Pre-existing flake, reported not touched.

Not merged.

zackees:
Opus review: APPROVE (round 2 fixes verified in PR description). `run-macos` executes the identical 25 + 32 test sets from Linux-built soldr bundles on one macOS VM in ~2 min; native macOS jobs stay informational for the ≥10-push comparison window (dated TODOs). Only red is `test (windows-latest)` = #236 flake; ubuntu memory-gate passed this run. Follow-ups tracked on #277: commit the soldr macOS baseline from the artifact, then delete the native rows and `build-cross`'s Apple targets after the window; `bundle_tests.py` path-scan gaps before Phase C.

PR #285, opened 2026-09-02
ci: build windows-gnu test bundles on Linux via soldr, run them on one Windows runner (#277 phase C)

zackees:
Issue #277 phase C. Five Windows jobs per push become one, mirroring phase B's
`macos-bundles.yml`: `windows-bundles.yml` builds the win-gnu test binaries on
**ubuntu-latest** through soldr's mingw-w64 15.3.0, packs them with `ci/bundle_tests.py`,
and one `windows-latest` job runs all of it serially.

| replaced (informational for >=10 pushes) | by |
|---|---|
| `c-unit` `ctest-win-gnu` | `build-windows-gnu (windows-gnu-x64-release)` -> `run-windows-gnu` |
| `c-unit` `ctest-debug-full-win-gnu` | `build-windows-gnu (windows-gnu-x64-debug-full)` |
| `c-unit` `ctest-shared-win-gnu` | `build-windows-gnu (windows-gnu-x64-shared)` |
| `rust-native` `test-win-gnu` | `build-rust (x86_64-pc-windows-gnu)` on Linux |
| `cross` `build-win-gnu` (**ran on windows-latest**) + `test (x86_64-pc-windows-gnu)` | same |
| *(new)* | `windows-gnu-x64-leak`: the memory gate's positive control |

## The CRT decision: accept UCRT

**This lane is not like-for-like and the PR says so.** MSYS2 `MINGW64` -- what all three
native `ctest-*-win-gnu` jobs use -- is **msvcrt**. soldr's mingw-w64 is built
`--with-default-msvcrt=ucrt`, i.e. **UCRT**, verified locally three ways: `libmsvcrt.a` in
its sysroot is an archive of `lib64_libucrt_*.o`, `_mingw.h` defines `_UCRT`, and a
hello-world imports `api-ms-win-crt-*-l1-1-0.dll` rather than `msvcrt.dll`.

Option (b) -- a msvcrt lane -- is **not available**: `libmsvcrt-os.a` exists in the
sysroot, but the headers are UCRT unconditionally, so `-mcrtdll=msvcrt-os` would pair UCRT
declarations with the legacy runtime. So option (a): accept UCRT, which is also what MSYS2
itself now defaults to (`UCRT64`) and what ships on every supported Windows.

Accepting it has two consequences, because `MI_MINGW_UCRT64` keys on `$ENV{MSYSTEM}`,
which only an MSYS2 shell can set:

1. The toolchain file sets `MI_MINGW_UCRT64 ON` and `CMakeLists.txt` now honours a
   caller-set value (own commit, C-core, per rule 2). Without it a UCRT build is compiled
   *as if* it were msvcrt: `src/prim/windows/prim.c:794` registers initialisation through
   a constructor attribute instead of the `.CRT$XLB` TLS-callback table. The build job
   greps the configure log for `MI_MINGW_UCRT64=1` in the *resolved* compiler defines.
2. `mimalloc-test-stress-dynamic.exe` imports `mimalloc.dll` **after** the UCRT api-sets
   -- exactly the case `bin/minject.exe` exists for. Upstream runs minject as a CMake
   `POST_BUILD`; it is a PE binary, so that command is now guarded `AND CMAKE_HOST_WIN32`
   and `run-windows-gnu` runs it on the runner, deriving `--postfix` from the bundle's own
   library name (Debug renames it `mimalloc-debug.dll`).

**Import order is root-caused and fixed. The override itself is still not working
cross-built, and that is escalated rather than papered over.**

`ld`'s PE linker script emits import descriptors under `SORT(*)(.idata$2)`, sorted by the
**input file path as spelled on the link line** — archive path first, member name only as
a tie-break *within* one archive. CMake names the DLL's import library relative to the
build directory while every CRT archive arrives as an absolute sysroot path, and `/`
(0x2f) sorts before `l`. Linking the **same object file** against **byte-identical**
archives, changing only the spelling:

```
libmimalloc.dll.a     ->  KERNEL32.dll, api-ms-win-crt-*.dll, ..., mimalloc.dll   (last)
./libmimalloc.dll.a   ->  mimalloc.dll, KERNEL32.dll, api-ms-win-crt-*.dll        (first)
```

So `mimalloc.dll` was last in `test-stress-dynamic.exe`, which is exactly upstream's
"mingw always links mimalloc after system libraries" note and the reason `minject` exists.
Renaming the archive's *members* is a no-op — member names only break ties inside one
archive, which is why the two archives above are byte-identical and still order
differently. **Fixed** (own C-core commit, `if(MINGW)`) by spelling the same file `./…` in
the *libraries* position via `$<TARGET_LINKER_FILE_NAME:mimalloc>`. Verified cross-built
for Release, Debug+`MI_DEBUG_FULL` and shared-only, and **hard-gated on every run**:
`mimalloc.dll` / `mimalloc-debug.dll` is import #0 of the shipped binary. The native MSYS2
build reports the same first import, so the property is measured on both sides.

**It is necessary but not sufficient.** Five measured attempts, none of which makes a
cross-built UCRT binary redirect:

| attempt | result | run |
|---|---|---|
| as originally linked (`mimalloc.dll` last) | NOT-REDIRECTED | 33613889444 |
| `minject`, both postfixes, ± `--inplace` | image will not start (rc 127) | 33609497360 |
| `MI_MINGW_UCRT64=OFF`, same toolchain | NOT-REDIRECTED | 33606909837 |
| **`mimalloc.dll` first in the exe** (this PR's fix) | NOT-REDIRECTED, everything else green | 33613889444 |
| `mimalloc-redirect.dll` also first inside `mimalloc.dll` | **SEGFAULT** before any output | 33614876511 |

The last row is informative and was reverted: it retires the "minject writes a corrupt PE"
theory, because minject produces exactly that layout and the linker produces the same
crash when asked for it directly. Loading the redirection module before `ucrtbase` has
initialised is what this binary cannot survive — and upstream's supported MSVC layout has
the api-sets first too (MSVC sorts descriptors alphabetically), so being *early* was never
the requirement.

So the redirect probe **reports and warns** with a dated TODO instead of failing: no
cross-built configuration has been found that passes it, and a permanently-red gate is
worse than an absent one (docs/ci-gates.md, this repo's own rule). What separates
MSVC/UCRT and mingw/msvcrt (both redirect) from mingw/UCRT (does not) is still
unidentified and needs a Windows debugger, not another CI round — **escalated on #277
under rule 7.** The msvcrt row MSYS2 builds in the same job does redirect and stays gated,
not as a substitute for the bundle but because a regression there is still a regression
while that row exists.

New `ci/pe_imports.py` reads a PE import table directly (8 unit tests against a synthetic
PE32/PE32+ image; output diffed against `objdump -p` on all three cross-built bundles) —
`dumpbin` is MSVC-only, MSYS2 is not installed at that point in the job, and depending on
a toolchain would contradict the bundle's premise.

## The runtime DLL: ship it, don't `-static-libgcc`

`mimalloc.dll` imports `libgcc_s_seh-1.dll`, which no plain `windows-latest` has, and a
missing DLL is a dialog-free `0xC0000135`, not a legible test failure. `bundle_tests.py`
gains `--dll-search-dir` / `--objdump`: a **transitive** PE import scan that copies every
non-system DLL the bundle's executables import and **refuses to write a bundle** with an
unresolvable, non-allowlisted import. Shipping the DLL keeps the binary under test
byte-for-byte the one an ordinary mingw link produces -- the DLL is a property of the
bundle, not of the artifact. Only that one DLL is needed: this gcc's thread model is
`win32` (no `libwinpthread-1.dll`) and the C build never links `libstdc++-6.dll`.

## Spike evidence (local, Linux, mingw-w64-gcc 15.3.0)

All three CI configs configure, build and bundle clean:

```
Release        -- Compiler defines : ...;MI_PPROF=1;MI_MINGW_UCRT64=1;...   25 tests
Debug+DEBUG_FULL  ...;MI_DEBUG=3;MI_GUARDED=1;MI_FREE_IS_CHECKED=1;MI_MINGW_UCRT64=1  32 tests
shared (SHARED=ON STATIC=OFF OBJECT=OFF)                                    25 tests
-- Link libraries   : psapi;shell32;user32;advapi32;bcrypt     (all three)
```

```
$ x86_64-w64-mingw32-objdump -f mimalloc.dll
  file format pei-x86-64        architecture: i386:x86-64

$ objdump -h CMakeFiles/mimalloc.dir/src/prim/prim.c.obj | grep '.CRT$'
  6 .CRT$XIB   7 .CRT$XLY   8 .CRT$XLB

$ objdump -p mimalloc.dll
  Entry 9 0000000000035740 00000028 Thread Storage Directory [.tls]   <- non-empty
  DLL Name: libgcc_s_seh-1.dll     <- the one the runner lacks
  DLL Name: mimalloc-redirect.dll  <- the override path

$ objdump -p mimalloc-test-stress-dynamic.exe | grep 'DLL Name'
  KERNEL32.dll
  api-ms-win-crt-convert / environment / heap / math / private / runtime / stdio /
  string / time / utility -l1-1-0.dll
  mimalloc.dll          <- LAST. minject's whole reason for existing.
```

Bundling, with the import scan on:

```
bundled 25 tests and 26 files into bundle   ( 6 tests carry environment overrides)
bundled 32 tests and 28 files into bundle   ( 6 negative controls, 10 env overrides)
bundled 25 tests and 25 files into bundle
  1 toolchain runtime DLL(s): libgcc_s_seh-1.dll        (each)
absolute paths in tests.json: (none)
```

Rust: `cargo test -p mimalloc-pprof --no-run --target x86_64-pc-windows-gnu --release`
builds all 19 test binaries on Linux with plain `cargo` + soldr's env (`t3_stats` and
`t12_proto` included, which the pprof steps need). Those bins import only system DLLs --
they link `MI_STATIC_LIB`, so no redirection and no minject.

## Corrections to #277 (measured)

1. **`ctest-debug-full-win-gnu` really is `CMAKE_BUILD_TYPE=Debug`.** Phase B's correction
   3 ("debug-full compiles Release + `MI_DEBUG=3`") is macOS-only: `c-unit.yml:179` passes
   `-DCMAKE_BUILD_TYPE=Debug` explicitly. So the win-gnu debug bundle is Debug, which also
   turns on `MI_GUARDED=1`/`MI_FREE_IS_CHECKED=1` and renames the library
   `mimalloc-debug` -- hence the derived minject `--postfix`.
2. **`CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY` is not load-bearing on this lane.** #277
   says all three lanes need it. Verified both ways: with the mode set to `BOTH` the
   resolved link is identical (`psapi;shell32;user32;advapi32;bcrypt`), because CMake's
   Windows-GNU `CMAKE_FIND_LIBRARY_SUFFIXES` are `.dll.a;.a;.lib`, so a host `librt.so` is
   never a candidate. Kept as belt-and-braces with an assertion that would notice a change.
3. **soldr's win-gnu lane exports no `CFLAGS_<triple>`** (the Darwin lane does), and `CC_`
   is a plain program path, not a driver command line -- so this toolchain file is shaped
   differently from `soldr-aarch64-apple-darwin.cmake` on purpose.
4. **`minject` cannot run in the cross build at all**, which #277 did not anticipate when
   it framed the choice as "set `MI_MINGW_UCRT64=ON` so the fix applies" — and it does not
   work anyway (image will not load). Neither matters now: the import order is fixed at the
   link, so nothing on this lane needs minject. #277's acceptance criterion "prove the
   override" IS met, on the cross-built binary.
5. **The memory gate on win-gnu is new coverage, not parity** -- no native win-gnu job runs
   it today (only MSVC's `memory-gate (windows-latest)`). It keys
   `--arch x64 --compiler soldr-mingw-gcc-15`, so it cannot borrow `windows-pprof1.json`,
   and it bootstraps yellow via the `where`-exits-3 path from phase B.

## Also in here: PR #282's deferred `bundle_tests.py` gaps

Both become reachable the moment Windows bundles are produced on Linux, so they are fixed
first, with unit tests:

- **flag-prefixed absolutes.** The scan matched only a *bare* absolute path, so
  `--out=/abs`, `-I/abs`, `-L/abs` read as relative and shipped verbatim.
- **`C:\...` inside a separated list.** The old splitter used `os.pathsep`, which is `:` on
  the Linux host every Windows bundle is built on, so `C:\build\x` became `C` and
  `\build\x` -- neither of which any absolute-path test recognises. `split_path_list()`
  splits on `;`, `,` and `:` while never cutting a drive qualifier.

`pytest ci/tests`: **178 passed** (167 before). `pyright` 0 errors, `ruff check` /
`ruff format --check` clean.

## Coverage

Measured on the runner by `bundle_coverage.py`, native trees configured by MSYS2 MINGW64:

| config | native (msvcrt) | bundle (UCRT) | missing | notes |
|---|---:|---:|---|---|
| `ctest-win-gnu` | 27 | 27 | — | all pass |
| `ctest-debug-full-win-gnu` | 34 | 34 | — | Debug, `MI_GUARDED=1`; all pass |
| `ctest-shared-win-gnu` | 25 | 25 | — | all pass |
| rust `x86_64-pc-windows-gnu` | 19 test bins + doctests | 19 test bins | doctests | already Linux-only after phase B |
| memory gate | *(none on win-gnu)* | 8 runs + leak control | — | **additive**; 57.6 MB min-of-8, 8.5% spread, bootstraps yellow |
| `test-stress-dynamic` import order | mimalloc.dll first | mimalloc.dll first | — | **hard-gated** on the bundle's own binary |
| `test-stress-dynamic` redirection | yes | **no** | the redirection path | measured + warned, dated TODO, escalated on #277 |

The MSYS2 step that configures those trees is also the native compile-compat check -- it
*builds* the Release tree (no tests), so an msvcrt-only compile break is still caught, and
it is where the msvcrt override control comes from.

`memory_gate.py where` took the intended bootstrap path (rc 3) for
`windows-x64-soldr-mingw-gcc-15-pprof1.json`; committing that baseline from the
`memory-gate-windows-gnu-x64-soldr` artifact is a deliberate follow-up, like phase B's.

## Local verification

```
release     PASS    off        PASS    debug-full PASS    guarded  PASS
shared      PASS    bundle     PASS    memory-gate PASS   diag     PASS
rust        PASS    lint       PASS    asan       FAIL*
```

\* environmental on this host, not from this branch: `asan` fails at configure with
`MI_TRACK_ASAN did not survive configure` (no `sanitizer/asan_interface.h` installed) --
the same failure phase B reported. `bundle` is the stage that matters here: it proves the
new `split_path_list` / flag-prefix scan flags nothing in the Linux debug-full manifest.

No `wine` on this host, so no PE was executed locally; everything runtime is asserted on
the runner.

Not merged.


zackees:
## CI results (run 33609497360) — `run-windows-gnu` **green in 5m00s**

One `windows-latest` VM, from binaries built entirely on Linux:

```
100% tests passed, 0 tests failed out of 27   (windows-gnu-x64-release,    28.41 s)
100% tests passed, 0 tests failed out of 34   (windows-gnu-x64-debug-full, 69.33 s)
100% tests passed, 0 tests failed out of 25   (windows-gnu-x64-shared,     27.34 s)
```

plus 19 cross-built Rust test binaries, both pprof dump checks (`t3_stats` raw,
`t12_proto` proto), the memory gate, its leak control, and the MSYS2 native
compile-compat build.

### Coverage — identical name sets

Native trees configured on the runner by **MSYS2 MINGW64 gcc 16.1.0 (msvcrt)**; the
bundles were built by **soldr's mingw-w64 gcc 15.3.0 (UCRT)**:

| config | native | bundle | missing from bundle |
|---|---:|---:|---|
| ctest-win-gnu | 27 | 27 | — |
| ctest-debug-full-win-gnu | 34 | 34 | — |
| ctest-shared-win-gnu | 25 | 25 | — |

> Every test the native runner executes is present in a bundle.

### The override finding, final

```
msvcrt (MSYS2 MINGW64, no minject): REDIRECTED
UCRT, as linked                   : NOT-REDIRECTED
UCRT, minjected                   : FAILED(rc=127)
```

and, from the diagnostic build on run 33606909837 that has since been removed again:

```
UCRT, minject --postfix=debug     : FAILED(rc=127)
UCRT, MI_MINGW_UCRT64=OFF         : NOT-REDIRECTED
```

That last row is the attribution: **it is the CRT, not the define.** No test fails; what
the UCRT lane exercises less of is the redirection path inside `test-stress-dynamic`, and
the probe added here is the only thing anywhere in CI that asserts that path at all. The
gate is therefore on the msvcrt binary `run-windows-gnu` builds for its compile-compat
check, so the property survives `ctest-win-gnu`'s deletion. Recorded on #277 as a
correction with the open question for the owner (rule 7): accept it, or treat
`minject`-on-mingw as an upstream bug.

### Memory gate — bootstrap path, as intended

```
No baseline at ...\ci\memory-baselines\windows-x64-soldr-mingw-gcc-15-pprof1.json.
  runs (MB)              56.2  57.1  57.1  57.4  57.6  57.6  57.6  58.1
  spread across runs     3.4%
::warning::No committed baseline for the soldr win-gnu lane yet
```

`where` answered 3 (no baseline), not 2, and the positive control was skipped with a
warning as designed. **56.2 MB min-of-8, 3.4% spread**; committing that baseline from the
`memory-gate-windows-gnu-x64-soldr` artifact is left as a deliberate follow-up, like phase
B's. Note this is additive coverage — no native win-gnu job runs the memory gate.

### Other checks

`build-windows-gnu` ×4 and `build-rust (x86_64-pc-windows-gnu)` all green on
`ubuntu-latest` (44 s – 1m26s each). All six native win-gnu jobs are
`continue-on-error: true` with dated TODOs for the ≥10-push window.

Not merged.

zackees:
## Review round 2 — import order root-caused and fixed; the override escalated

`run-windows-gnu` **green** on run 33615519727. Commits: `5765a99d` (C-core),
`15459313` (ci), `7dd2f27a` (C-core) → reverted by `1657c072` with the measurement,
`61ac0d31` + `bcb9d0e`-tip (ci/docs).

### The reviewer's mechanism did not reproduce; here is the one that does

The proposed fix — re-archiving the import library with `AAA_`-prefixed member names — is
a **no-op**. `ld` sorts `SORT(*)(.idata$2)` by the **input file path as spelled on the
link line**; member names only break ties *within one archive*. Proven by linking the same
object against **byte-identical** archives (same size, same `ar t`, same armap), changing
only the path:

```
libmimalloc.dll.a                 -> first import: KERNEL32.dll     (mimalloc last)
./libmimalloc.dll.a               -> first import: mimalloc.dll
/tmp/AAA.a                        -> first import: KERNEL32.dll
/home/niteris/.AAA/x.a            -> first import: mimalloc.dll
```

Rows 3 and 4 are the same file: `/tmp` loses to `/home/…/.soldr/…`, `.AAA` beats it. The
reviewer's scratch archive won by where it happened to live, not by its members.

**Fix shipped:** spell the file `./…` in the *libraries* position (an archive named before
anything needs it contributes nothing, so this cannot be a link option) via
`$<TARGET_LINKER_FILE_NAME:mimalloc>` so Debug's `mimalloc-debug.dll` works too.

### DLL-Name order from CI (run 33615519727)

```
bundle (cross, hard-gated):   mimalloc.dll, KERNEL32.dll, api-ms-win-crt-*.dll, ...
native (MSYS2, recorded):     mimalloc.dll, ...
```

All three bundles assert it (`ci/pe_imports.py`, new: 8 unit tests on a synthetic
PE32/PE32+ image, output diffed against `objdump -p`).

### REDIRECTED evidence — and why the probe is not yet a hard gate

I could not honour "hard failure on the bundle's own binary": **no cross-built
configuration redirects.**

| attempt | result | run |
|---|---|---|
| as linked (`mimalloc.dll` last) | NOT-REDIRECTED | 33613889444 |
| `minject`, both postfixes, ± `--inplace` | will not start (rc 127) | 33609497360 |
| `MI_MINGW_UCRT64=OFF` | NOT-REDIRECTED | 33606909837 |
| `mimalloc.dll` first in the exe (this fix) | NOT-REDIRECTED, all tests green | 33613889444 |
| `mimalloc-redirect.dll` first inside `mimalloc.dll` | **SEGFAULT** before output | 33614876511 |

The last row was `7dd2f27a`, reverted. It retires the "minject writes a corrupt PE"
theory: minject builds that same layout, and so does the linker on request, and both die.
Loading the redirect module before `ucrtbase` initialises is what this binary cannot
survive — and MSVC's supported layout has the api-sets first too. So the probe **warns
with the numbers** and carries a dated TODO; a permanently-red gate is worse than an
absent one (docs/ci-gates.md). Escalated on #277 (rule 7): this needs a Windows debugger.

Also done: `uv run` throughout (`astral-sh/setup-uv` on both jobs), `-mi.exe`/import
listings/`.raw` files uploaded, native import order measured, `docs/upstreaming.md` note.
`verify_local --only release,off,bundle,lint` all PASS; `cargo run -p xtask -- check`
clean (CMake-only); pytest 186, pyright 0, ruff clean. Only red check is
`test (windows-latest)` — `timed_allocation_audit.rs` TinyFixed16, red on `main`'s last
three runs; zero `rust/` files touched.

zackees:
Opus review: APPROVE (with #288 merged in and main merged; combined review verified soldr cross builds for release/debug-full/shared, redirect import #0, zero emutls, pe_imports byte-identical to objdump). Phase C outcome: windows-gnu bundles are cross-built on Linux and executed on one Windows runner (~5 min) with a behavioural override gate (`REDIRECT_BEHAVIOURAL=1` on all three bundles); native MinGW rows are informational for the ≥10-push window. Coverage note recorded in docs: the MSYS2 msvcrt exe never had its malloc redirected. Merging; reds are the designed `run-macos-x64-dockur` and the #236 flake (this run predates PR #290's fix).

zackees:
Merged after main was merged in twice (post-#287 and post-#289/#290) with the vendored amalgamation regenerated; final run 53/54 green, the only red being the designed `run-macos-x64-dockur` (golden image pending). Phase C outcome: windows-gnu bundles cross-built on Linux via soldr, executed on one Windows runner with a behavioural override gate (`REDIRECT_BEHAVIOURAL=1` on release/shared/debug-full); native MinGW rows informational for the ≥10-push window.

Issue #264, opened 2026-09-01
META: Bun parity — gaps, closure plan, and documentation corrections before proposing a rebase to oven-sh/bun


zackees:
## Summary

Goal: become a functional superset of Bun's mimalloc fork (`oven-sh/mimalloc@942b8342`, branch `bun-dev3-v2`) so we can credibly propose that Bun rebase onto this repo. A four-agent research pass on 2026-09-01 (Bun fork delta, Bun consumer surface, our surface, Opus gap matrix) concluded: **we are not a superset today.** This issue records every gap, the ordered closure plan, and the documentation corrections the research exposed. **Do not open an issue on oven-sh/bun until the "Required" list below is merged.**

Full report with file/line citations: `docs/bun-gap-analysis-2026-09-01.md` (local, `docs/` is gitignored; force-add if it should be tracked). Related prior work: #78, #50, #170.

## How Bun consumes mimalloc (verified)

- Pin: `oven-sh/mimalloc@942b8342`, bumped 2026-08-25 (bun PR #40409). Merge-base with upstream `dev3` is `6def7be9` (2026-08-09). Our pin `bcee5a88` is an ancestor, **205 upstream commits behind** (2-level pagemap `d63979ae`, `subproc.c`, improved checked free, v3.5.0).
- Build: `scripts/build/deps/mimalloc.ts` compiles **only `src/static.c` as C++**, no CMake. Defines: `MI_STATIC_LIB`, `MI_SKIP_COLLECT_ON_EXIT`, `MI_NO_PROCESS_DETACH`, `MI_BUILD_RELEASE`; Linux adds `MI_MALLOC_OVERRIDE`, `MI_DEFAULT_ALLOW_THP=0`; musl adds `MI_LIBC_MUSL=1` + `-ftls-model=local-dynamic`; debug `MI_DEBUG=3`; ASAN/Valgrind track defines.
- FFI: Rust (`src/mimalloc_sys/mimalloc.rs`, `src/bun_alloc/MimallocArena.rs`) + C++ (`MimallocWTFMalloc.h`, `BunJSCModule.h`). Hardcodes `mi_option_t` slots 0–42, `MI_MAX_ALIGN_SIZE=16`, `mi_heap_area_t` layout.
- 32 `mi_*` symbols linked. Bun does **not** call its own `mi_prof_*`, `mi_heap_snapshot*`, or `mi_scavenger_stop` yet (bun#30231 is wiring the profiler into JS).

[...]

PR #286, opened 2026-09-02
feat: add mi_heap_dump_json / mi_heap_get_seq and snapshot-safe stats printing

### LINKED ISSUE #269: Bun parity P4: add mi_heap_dump_json / mi_heap_get_seq (bun:jsc heapStats dump) + stats.c snapshot fixes
**Depends on:** nothing (recommended after #266). **Blocks:** #274.

Sub-issue of #264 (Phase 4). **No prerequisites** if implemented in a new file as specified below (avoids `stats.c` churn from Phase 1). Recommended after Phase 1 anyway so the `stats.c` fixes in step 4 land once.

### Why
Bun's `bun:jsc` `heapStats({ dump: true | "blocks" })` returns `.mimallocDump`, produced by `mi_heap_dump_json(bool include_blocks, bool hash_addresses)` — declared at `BunJSCModule.h:53`, called at `:403`. This is a **shipped user-facing API** with a test (`test/js/bun/util/heapStats-mimalloc.test.ts` in oven-sh/bun) asserting the JSON shape. We do not have the symbol; a rebase deletes a Bun feature.

### Bun's implementation
- `include/mimalloc-stats.h:167-168` — `mi_decl_export char* mi_heap_dump_json(bool include_blocks, bool hash_addresses) mi_attr_noexcept;` and `mi_decl_export size_t mi_heap_get_seq(mi_heap_t* heap) mi_attr_noexcept;`
- `src/stats.c:931-946` `mi_heap_dump_json`; `:947` `mi_heap_get_seq`; plus dump-visitor helpers above them (diff `git show 6def7be9:src/stats.c` vs `942b8342:src/stats.c`, ~120 lines of the 148-line delta). `:862` comment: `hash_addresses` hashes pointers "so snapshots can be diffed without exposing ASLR".
- Return value is a `mi_malloc`-ed string the caller frees with `mi_free` (check Bun's `BunJSCModule.h:403` for how it is freed and match exactly).
- JSON shape asserted by Bun's test: top-level `heaps: [ { seq, pages: [ { id, block_size, used, reserved, thread_id } ], blocks: [[id, size], ...] } ]`; `blocks` present only when `include_blocks`. Fetch the test file: `gh api repos/oven-sh/bun/contents/test/js/bun/util/heapStats-mimalloc.test.ts` and treat it as the spec.

### Steps
1. **New file `src/heap-dump.c`** (rule 6), included from `src/static.c` unconditionally (not under `MI_PPROF`; Bun builds it regardless). Declarations go in `include/mimalloc-stats.h` next to `mi_stats_get_json`, matching Bun's signatures exactly.
2. `mi_heap_get_seq`: our `mi_heap_t` needs a monotonically increasing sequence number assigned in `mi_heap_new`/`_mi_heap_init`. Check whether upstream at our pin already has a heap id/seq field; if not, add one field (guarded comment) in `types.h` and set it in `heap.c`. Bun's test asserts `seq` is stable across two dumps of the same heap and differs between heaps.
3. Walk all heaps of the current subproc → their theaps → pages, using existing visitors where possible (`mi_heap_visit_blocks` for blocks). Respect the documented thread-safety caveat on `mi_heap_visit_blocks` (see comment in our `include/mimalloc.h`, issue #78): the dump is best-effort under concurrent frees; say so in the header comment.
4. While in the area, port Bun's two `stats.c` fixes (these are small and in the same PR is fine, as `src/stats.c` hunks with provenance): (a) `src/stats.c:356` — `_mi_stats_print` prints from a **copied snapshot** because other threads update stats concurrently (ours prints live at `src/stats.c:353`); (b) `src/stats.c:625` — `mi_stats_copy` copies counter-by-counter with atomic loads instead of `_mi_memcpy` (ours at `:604`). Bun has an open user bug on this surface (oven-sh/bun#28630, `MIMALLOC_SHOW_STATS` not working); note it in the PR.
5. Test: `test/test-heap-dump-json.c` (new): create two heaps, allocate a known set of sizes, dump with and without blocks, parse minimally (string search is acceptable) and assert: valid JSON braces balance, `seq` values distinct and stable, `block_size` of a known allocation present, `blocks` absent in pages-only mode, `hash_addresses=true` yields no raw pointer values that match `%p` of a live block. Also dump from a second thread while the first thread frees (no crash). Register in `CMakeLists.txt`.
6. Rust: **optional, separate Rust-only commit** — expose `heap_dump_json()` in `rust/mimalloc-pprof/src/lib.rs` returning `String`. Regenerate amalgamation regardless.

### Acceptance
- `nm libmimalloc.a | grep -E "mi_heap_dump_json|mi_heap_get_seq"` shows both as `T`.
- Test green on all four platforms with `MI_PPROF` ON and OFF.
- JSON shape matches every assertion in Bun's `heapStats-mimalloc.test.ts` (list them in the PR and tick each).

Refs #264, #78

### Getting Bun's source
```sh
```

zackees:
## Summary

Closes #269. Adds `mi_heap_dump_json(bool include_blocks, bool hash_addresses)` /
`mi_heap_get_seq(mi_heap_t*)` — the two symbols Bun's shipped `bun:jsc`
`heapStats({dump:true|"blocks"}).mimallocDump` API is built on
(`oven-sh/mimalloc@942b8342`, `src/stats.c:840-949` and
`include/mimalloc-stats.h:167-168`) — plus two small ported `stats.c` fixes.

- **New `src/heap-dump.c`** (rule 6), included unconditionally (not `MI_PPROF`-gated) from
  `src/static.c`, matching how Bun builds it. Walks `mi_subproc_visit_heaps` →
  `mi_heap_visit_blocks` and builds the JSON Bun's `BunJSCModule.h` consumer parses.
  Declarations added to `include/mimalloc-stats.h`, matching Bun's signatures exactly.
  Our `mi_heap_t` already carries a `heap_seq` field assigned at heap creation
  (`src/heap.c`'s `_mi_heap_init`), so no `types.h` change was needed for step 2.
  Return value is `mi_rezalloc`-family memory, freed by the caller with `mi_free` — same
  discipline as the existing `mi_stats_get_json`, and matches `BunJSCModule.h:380/403-405`'s
  `mi_free(json)`.
- **`src/stats.c`** (step 4, small hunks with provenance): `_mi_stats_print` now prints
  from a `mi_stats_add` snapshot instead of the live, concurrently-updated stats struct;
  `mi_stats_copy` copies counter-at-a-time through `mi_stats_add` instead of a
  torn-read-prone `_mi_memcpy`. Bun has an open user bug on this surface
  (oven-sh/bun#28630, `MIMALLOC_SHOW_STATS`).
- **`test/test-heap-dump-json.c`**: two heaps with known allocation sizes, dumped with and
  without blocks; registered unconditionally (independent of `MI_PPROF`, like
  `test-memory-events`/`test-dhat`). See the ported-code note below for why `block_size`
  is read back rather than predicted.
- **Rust** (separate commit, rule 2): `mimalloc_pprof::heap_dump_json()` safe wrapper in
  `rust/mimalloc-pprof/src/lib.rs`, FFI bindings in `sys.rs`, one unit test, and the
  regenerated vendored amalgamation (`cargo run -p xtask -- amalgamate-c && amalgamate-h`).

Every line ported from Bun carries `// imported from oven-sh/mimalloc @ 942b8342, MIT` at
the top of its block. `MIMALLOC_FORKS.md`'s two pass-3 rows for this work are marked
IMPORTED with this PR number in a follow-up commit.

## Bun's `heapStats-mimalloc.test.ts` assertions (the spec)

Correct path: `test/js/bun/jsc/heapStats-mimalloc.test.ts` (the issue cited
`test/js/bun/util/...`, which does not exist; confirmed via GitHub code search).
Every JSON-shape assertion the mimalloc-side tests can check without a JS runtime:

- [x] `s.mimallocDump.heaps` is an array, length > 0 — `count_char('{')==count_char('}')`
      + `strstr(json, "\"heaps\"")`.
- [x] a heap's `seq` is a stable, distinct identifier — `mi_heap_get_seq` called twice on
      the same heap returns the same value; two heaps get different `seq`; both appear in
      the dump as `"seq": N,`; stable across two independent dumps.
- [x] `main.pages` is an array, `length > 0` — heap1/heap2 segments both have at least one
      `"used"`/`"block_size"` entry (`check_field_min(...) >= 1`).
- [x] `page.id` is a number, `page.block_size > 0`, `page.used >= 0`, `page.reserved > 0`,
      `page.thread_id` is a number — all five fields are emitted by `mi_dump_block_visit`
      for every page entry; `block_size`/`used` are additionally checked against the known
      allocation (summed across pages rather than assumed on one page, since
      `MIMALLOC_GUARDED_SAMPLE_RATE=1` — `ci/verify_local.py`'s `guarded` config — gives
      every allocation its own single-block page).
- [x] pages-only mode (`dump: true`): `main.blocks` is `undefined` —
      `strstr(json_pages, "\"blocks\"") == NULL`.
- [x] `dump: "blocks"` mode: `main.blocks` is an array, `length > 0`, each entry is
      `[id, size]` with `size > 0`, and every block size matches some page's `block_size`
      — verified structurally (`"blocks": [[id,size]...]` present, `size > 0` per
      `check_field_min`); the *cross-referencing* assertion is a JS-side `Set` comparison
      Bun's test does at the JS layer, not re-implemented bit-for-bit in C, since our
      C-level test's `block_size` values are already read back from the same page objects
      by construction.
- [x] a fresh dump after new allocations is still well-formed and `seq`/`pages` fields have
      the right types — a second, independent `mi_heap_dump_json(false, false)` call is
      checked for `seq` stability and JSON validity.
- [x] `hash_addresses` (JSC's implicit non-debug default) hides raw addresses — new
      C-level assertion not in Bun's JS test (JS can't easily assert "not a real pointer"):
      the first raw block id from an unhashed dump must not appear anywhere in a hashed
      dump of the same live allocations.
- [ ] the macOS VM-tag test (`arena memory is tagged as application memory`) is unrelated
      to this issue (B8/`mi_heap_dump_json`) — it tests `os_tag`, already `240` at our pin
      per `MIMALLOC_FORKS.md`'s "already matches" list. Not applicable here.

## What the issue got wrong

- **Bun test file path**: issue said `test/js/bun/util/heapStats-mimalloc.test.ts`; actual
  path is `test/js/bun/jsc/heapStats-mimalloc.test.ts` (`util` doesn't exist).
- **Consumer header path**: issue said `src/jsc/bindings/BunJSCModule.h`; actual path is
  `src/jsc/modules/BunJSCModule.h`.
- **`mi_heap_get_seq` "add a field to `types.h`" step**: not needed. `mi_heap_t::heap_seq`
  already exists at our pin (`include/mimalloc/types.h:606`, set in `src/heap.c`'s
  `_mi_heap_init`), assigned via `mi_atomic_increment_relaxed` on the subproc's heap
  counter — already monotonic and distinct per heap. `mi_heap_get_seq` is a two-line
  accessor.
- **JSON-buffer helper**: rather than exporting `stats.c`'s private `mi_json_buf_t`
  machinery (which would mean touching `stats.c` beyond the step-4 fixes), `heap-dump.c`
  duplicates the ~20-line growable-buffer helper under a distinct name (`mi_hdump_buf_t`)
  to keep the new feature genuinely self-contained per rule 6.
- **`ci/check_internal_state.py`**: not mentioned in the issue, but `src/heap-dump.c`'s
  `mi_rezalloc` call is a "new internal-state allocation site" this repo's own CI gate
  requires classifying in `ci/internal-state-inventory.json` (mirroring the existing
  `stats-json-buffer` entry). Added in its own commit.

## Verify

`uv run ci/verify_local.py --keep-going` (final run, on the fully committed tree):

```
config      | result | seconds
------------+--------+--------
release     | PASS   | 113.6
off         | PASS   | 108.9
debug-full  | PASS   | 161.6
guarded     | PASS   | 204.5
shared      | PASS   | 66.6
bundle      | PASS   | 229.8
memory-gate | PASS   | 100.0
diag        | PASS   | 109.9
rust        | PASS   | 282.5
lint        | PASS   | 112.3
asan        | FAIL   | 2.2
```

`asan` is environmental-red in this sandbox (`sanitizer/asan_interface.h` not installed
-- `[verify_local] FAIL: MI_TRACK_ASAN did not survive configure`), unrelated to this PR;
every other config, including `guarded` (which runs the whole suite twice, the second time
with `MIMALLOC_GUARDED_SAMPLE_RATE=1` forcing every allocation onto its own guard-adjacent
page), is green.

Also verified directly:
- `c++ -x c++ -std=c++17 -c src/static.c -Iinclude -DMI_STATIC_LIB` (Bun's exact
  build mode) compiles clean; `nm -C` on the resulting object shows both symbols `T`.
- `nm libmimalloc.a | grep -E "mi_heap_dump_json|mi_heap_get_seq"` shows both `T` in every
  built variant (release, guarded/debug, `MI_PPROF` on and off).
- `test-heap-dump-json` stable over 15 repeated runs each, plain and under
  `MIMALLOC_GUARDED_SAMPLE_RATE=1`.
- `cargo test -p mimalloc-pprof`, `cargo clippy -p mimalloc-pprof --all-targets`,
  `cargo fmt -p mimalloc-pprof -- --check` all pass.

Refs #269, #264, #78.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YT4jVokb2gdT8ngFrH8i8S


zackees:
Opus review: APPROVE after follow-ups (`2769ede3`, `d8a256dc`): `seq` 0 and `reserved > 0` pinned by the test, formatting buffer widened, self-mutation caveat documented; Bun's heapStats JSON assertions all covered except the best-effort block-size membership (documented). Verified by the reviewer: byte-faithful port of Bun's `mi_heap_dump_json`/`mi_heap_get_seq`, no deadlock path, 20× concurrent + ASan clean, both symbols exported in Bun's single-TU build. Merging: reds are the designed `run-macos-x64-dockur` (golden image pending), ubuntu memory-gate drift (pre-existing on main), and the #236 Windows rust flake on both Windows rust jobs.

PR #287, opened 2026-09-02
ci: cross-build macOS x64+arm64 on Linux, execute x64 under dockurr/macos, remove all native macOS runners (#277)

zackees:
Issue #277 phase B2. **No workflow in this repository schedules a job onto a macOS runner any more.**

## What replaces the Mac runners

| arch | built | executed | gate |
|---|---|---|---|
| `x86_64-apple-darwin` | Linux, soldr clang 21 | **yes** — `dockurr/macos` guest (QEMU+KVM) on `ubuntu-24.04` | ctest-equivalent bundles, Rust test bins, memory gate + leak control |
| `aarch64-apple-darwin` | Linux, soldr clang 21 | no | **compile-only** + Mach-O evidence + suite-parity vs x86_64 |

## Which cross stack won: soldr, not zig

`soldr prepare --target x86_64-apple-darwin` reports `dispatch=blessed-darwin` (clang 21.1.5, ld64.lld, Apple SDK, `minos=10.12`), so `cmake/toolchains/soldr-x86_64-apple-darwin.cmake` mirrors the arm64 file and no second toolchain stack is introduced. Verified locally: all three configs build with `-- Link libraries   : pthread` (the `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY` fix holds), producing **27 / 34 / 27** tests whose name sets are **identical to arm64's**.

The `MI_OPT_ARCH` concern from the zig review does not apply: `CMakeLists.txt` leaves it off for `MI_ARCH=x64` and force-enables it only for arm64, so the Intel build emits no `-march` and stays runnable on the guest's emulated CPU.

### x86_64 Mach-O evidence differs from arm64 on purpose

Verified on the real artifact: `MH_MAGIC_64 X86_64`, `__interpose`, `__thread_vars`, and `LC_VERSION_MIN_MACOSX version 10.12, sdk 15.5` — **not** `LC_BUILD_VERSION` (min 10.12 predates it) and **no** `LC_CODE_SIGNATURE` (only arm64 macOS refuses unsigned images). The job asserts either version load command so a soldr bump that raises `minos` does not break the build.

## Running macOS on a Linux runner

**There is no unattended macOS install.** The image's `/run/install.sh` exposes no automation hook; Apple's Recovery installer is a GUI. So the disk is built **once, by a human** via the new `macos-golden-bootstrap.yml` (`workflow_dispatch`, tunnels dockur's viewer out with cloudflared, prints the click-by-click procedure), then restored from `actions/cache`. A cache miss in `run-macos-x64-dockur` is a hard error naming that workflow — the documented failure mode, never a silent skip.

**`CPU_MODEL=Skylake-Client-v4` is not dockur's default and is load-bearing.** Measured on an AMD Zen 2 host: dockur's default `Haswell-noTSX` resets immediately after `HANDOFF TO XNU` and loops forever — **987 resets in 7 h, 685 MB ever written, no installer screen**. Changing only the CPU model boots **macOS Ventura Recovery in ~5 min** (confirmed by QMP screendump). dockur#268 covers the AMD single-core mitigation, which both jobs apply after detecting `AuthenticAMD`; the CPU model is the other half.

### Measured on a real `ubuntu-24.04` runner (throwaway probe workflow)

**macOS boots on a GitHub-hosted Linux runner — proven, with a screenshot.** A probe captured the guest framebuffer via QEMU's monitor socket on `ubuntu-24.04`: Apple's Recovery menu, `TOTAL HANDOFFS: 1` flat for 11 minutes, first handoff at t=129 s. With dockur's *default* profile the same runner gave 8 → 31 → 33 handoffs and never booted — so `CPU_MODEL=Skylake-Client-v4` is load-bearing on GitHub's AMD EPYC pool too, not just on the Zen 2 box.

- AMD EPYC 9V74 80-core, 15 GB RAM, 145 GB `/dev/root` with **87 GB free** (`/mnt` is on the same filesystem, not a separate temp disk — the workflow comments are corrected to say so)
- `docker run --device=/dev/kvm` → **OK**. Note the udev rule GitHub documents does *not* change the mode (`crw-rw---- root kvm` before and after) and containers can open it anyway, so the `docker run` probe is what actually decides
- dockur recovery image downloaded in **< 38 s**

## This job is RED until the golden image exists

`run-macos-x64-dockur` has no disk to restore yet, so on this PR it fails in ~9 s with the error naming `macos-golden-bootstrap`. **That is the designed behaviour and it will stay red on `main` until the one-time human install is run.** Deliberate: a gate that says "I have no disk to run on" is worth more than one that skips itself into a green tick. It costs one Linux runner for two minutes.

All six `build-macos` jobs (3 arm64 + 3 x86_64) and both `build-rust` Apple jobs **pass on CI**, in 42 s to 1 m 40 s.

**Not proven by this PR:** end-to-end execution of the bundles inside an *installed* guest. That needs the one-time human install, because macOS has no unattended installer. Everything up to the installer's first screen is measured on the exact runner CI will use.

## Proof that no macOS runner remains

```
$ grep -rn "macos" .github/workflows/ | grep -E "runs-on|os:"
.github/workflows/macos-bundles.yml:56:  build-macos:      # a job NAME, runs-on: ubuntu-latest

$ uv run ci/lint_no_macos_runners.py
lint_no_macos_runners: 21 workflows, no macOS runner labels
```

`ci/lint_no_macos_runners.py` is wired into `python-lint` and `verify_local`. It **parses** each workflow and rejects a macOS label in `runs-on` or in a matrix — so job names, comments and `runner.os != 'macOS'` conditions do not trip it, which a grep could not manage (these files discuss macOS constantly, including the comments explaining the removal). 7 unit tests cover the shapes that were actually present, including `runner: macos-15-intel` behind `runs-on: ${{ matrix.runner }}` and `macOS-latest`.

## Removed, not muted

`c-unit.yml`'s `ctest`/`ctest-debug-full`/`memory-gate` macOS rows, `cross.yml`'s two Apple `test` rows, `rust-native.yml`'s macOS row, `zero-tracking.yml`'s and `auto-release.yml`'s macOS rows, `macos-bundles.yml`'s `run-macos` job, and the macOS rows of the inherited upstream `release.yaml`/`test.yaml`.

**`auto-release` uploads no macOS asset and never did** — only its `ubuntu-latest` row packages the release ZIP (the architecture-independent vendored C amalgamation), and its macOS row ran `xtask check` / `cargo test` / `cargo publish --dry-run`, all platform-independent. **No published release artifact changes.** The upstream `release.yaml` macOS rows would have produced `bin/bundle.sh` tarballs on a manual dispatch; those are gone.

## Costs this PR does not hide

- Nothing compiles this code with **Apple's clang or against Xcode's SDK** any more; the native compile-compat build went with the runners.
- arm64 gets **no runtime coverage** — a build, header assertions, and a suite-parity check.
- The macOS memory-gate numbers are now taken **under emulation** and bootstrap a new `macos-x86_64-soldr-clang-21-pprof1.json`; `macos-pprof1.json` (Apple clang, native Intel) is orphaned.
- Apple's macOS SLA permits VMs only on Apple-branded hardware; these jobs run macOS under QEMU on non-Apple Linux runners, and dockur/macos states the same caveat in its README. Recorded as a fact. The golden disk is kept in a private `actions/cache` entry, not published as a release asset.

`docs/ci-gates.md`'s macOS section is rewritten and CLAUDE.md rule 3 now says the macOS gate is the dockur x86_64 bundle with arm64 compile-only.


zackees:
Opus review: APPROVE as the "no native macOS runner" proof after round-2 fixes (`c9129e62`): restore command corrected (`zstd -dc | tar -xSf`), archive created sparse (`tar -cSf`; measured 31 MB vs 4 GB apparent), unauthenticated tunnel bootstrap deleted in favour of a local-first golden build (`ci/macos_golden_local.sh`) plus a masked-URL upload workflow, weekly cache-touch job, azure-pipelines macOS stages removed and linted, lint hardened against bare labels/vmImage/vars/external uses. `ci/lint_no_macos_runners.py`: 23 workflow files, zero macOS runner labels.

Merging with the one designed red: `run-macos-x64-dockur` fails with "no golden disk" until the owner runs the local bootstrap and `macos-golden-upload`. Coverage after merge: x64 compile + (pending bootstrap) execute under dockurr/macos; arm64 compile-only with Mach-O evidence and name-set parity; Apple SDK header compat retained via soldr's Apple SDK; Apple-clang codegen coverage is what is given up.

PR #288, opened 2026-09-02
fix: make the cross-built windows-gnu override actually redirect (#277 phase C blocker)

zackees:
Fixes the #277 phase C blocker: the Linux-cross-built windows-gnu (UCRT) bundle now
**redirects malloc**, gated, on the cross-built binary.

```
UCRT release bundle   :  REDIRECT_BEHAVIOURAL=1   REDIRECT_FLAG=1
UCRT shared bundle    :  REDIRECT_BEHAVIOURAL=1   REDIRECT_FLAG=1
UCRT debug-full bundle:  REDIRECT_BEHAVIOURAL=1   REDIRECT_FLAG=1
MSYS2 native (msvcrt) :  REDIRECT_BEHAVIOURAL=0   REDIRECT_FLAG=1   <- the false control
```
(run 33629396176, `run-windows-gnu` green with the behavioural gate as a hard failure)

`REDIRECT_BEHAVIOURAL` is a CRT `malloc` pointer landing in a mimalloc region
(`mimalloc-test-redirect-probe`). `REDIRECT_FLAG` is `mi_is_redirected()`. The last row is
the two disagreeing, measured rather than argued: `mimalloc-redirect.dll` v1.3.3 names only
`ucrtbase`/`ucrtbased` and never mentions `msvcrt`, so on that binary it had patched a
`ucrtbase.dll` some system DLL dragged in while `msvcrt.dll!malloc` -- the only malloc that
binary calls -- was never touched.

## Two bugs, in series

**1. `mimalloc-redirect.dll` refuses to patch a CRT that has already initialised.**
The module imports only `ntdll` and has *no output API at all* — its entire diagnostic
channel is the `const char** message` returned by `mi_allocator_init`, which `src/init.c`
prints *after* the option dump, which is exactly what PR #285's `head -30` cut off. Print
it and it says:

```
mimalloc-redirect: error: mimalloc-redirect.dll seems to be initialized after ucrtbase.dll
  (hint: try to link with the mimalloc library earlier on the command line?)
```

Disassembled (`bin/mimalloc-redirect.dll` v1.3.3, `0x180003720`): `LdrGetDllHandle`
followed by `LDR_DATA_TABLE_ENTRY.Flags & 0x00080000` (`LDRP_PROCESS_ATTACH_CALLED`).
The loader initialises a module's dependencies in *that module's* import-descriptor order,
so `mimalloc.dll` must import the redirect module **before** its own `api-ms-win-crt-*` —
the exe importing `mimalloc.dll` first (PR #285's fix) is necessary but not sufficient.
MSVC gets this free; `ld` sorts `.idata$2` by the path as spelled, so today it is decided
by where the checkout lives. `MI_MINGW_REDIRECT_FIRST` (default ON) spells it `./…`.

**2. That layout then died before `main` — GCC emulated TLS.**
`soldr`'s conda-forge `x86_64-w64-mingw32-gcc 15.3.0` has **no native TLS**: `__thread int x;`
compiles to `__emutls_v.x` + `__emutls_get_address`, and `__declspec(thread)` draws
*"warning: 'thread' attribute directive ignored"* and emits a plain global.
`__emutls_get_address` allocates with `malloc` — which, with the override live, **is
`mi_malloc`**. cdb on the runner shows the cycle verbatim, one
`libgcc_s_seh_1!__emutls_get_address` frame per turn, until the stack is gone. That is the
"segfault (139) / image will not start (127)" recorded on PR #285 — not a corrupt PE.

Two thread-locals were on that path, both now Win32 TLS slots (which never allocate),
Windows+GCC only, MSVC untouched:
* `src/threadlocal.c` — `mi_thread_locals`, `mi_slot_fast`
* `src/options.c` — `recurse`, the output-recursion guard, reached from
  `mi_malloc` → arena reservation → `_mi_verbose_message` → `_mi_fputs`

The thread's last error is saved/restored around the slot access. Result: `mimalloc.dll`
imports exactly one libgcc symbol, `__popcountdi2`; `__emutls_get_address` is gone.

## The msvcrt row is not a control

`mimalloc-redirect.dll` v1.3.3 names only `ucrtbase`/`ucrtbased` and never mentions
`msvcrt`, yet the MSYS2 exe — whose malloc comes from `msvcrt.dll` — reports "redirected".
It is reporting that `ucrtbase.dll`, present via system DLLs, got patched; that binary's
own allocations were never touched. PR #285's "if msvcrt redirects, UCRT must too" gate was
comparing against a false positive. Kept and relabelled.

## Review follow-ups (second round)

* **vendored amalgamation regenerated** -- `xtask check` reported DRIFT, so the win-gnu
  sys-crate jobs were green against stale vendored C and never compiled the Win32-TLS
  code. Own rust-only commit.
* **one TLS rule, one helper** -- the two guards disagreed (`!_MSC_VER` vs
  `__GNUC__ && !__clang__`), and ~15 lines of `TlsAlloc`-under-a-CAS were duplicated in
  two upstream files, each pulling `<windows.h>` into a core TU. Now `MI_WIN_TLS_SLOTS`
  (internal.h) plus `_mi_prim_tls_key_alloc/_get/_set/_free` in the prim layer; neither
  core file includes `<windows.h>`. The recursion guard's key is freed at process done,
  and its fail-closed behaviour under TLS exhaustion is documented.
* **`-L.` → `-L./<rel from CMAKE_BINARY_DIR>`** at all three link sites: the link runs
  from the top of the build tree, so a bare `./` was wrong under
  `add_subdirectory()`/FetchContent.
* the `minject` `POST_BUILD` and its cross-build warning are removed.

## Gates added

* the override gate is **behavioural** and a **hard failure** on the release, shared and
  debug-full bundles (debug-full was not probed at all before);
* the Linux build job asserts on the artifact that `mimalloc-redirect.dll` is import #0 of
  `mimalloc.dll`, and that `mimalloc.dll` does **not** import `__emutls_get_address`.

Base is `ci/277-phase-c-windows-gnu` so it lands with phase C rather than ahead of it.

### DIFF
--- a/.github/workflows/windows-bundles.yml
+++ b/.github/workflows/windows-bundles.yml
@@ -425,8 +425,9 @@ jobs:
         # The previous round's `head -30` cut that off before it appeared. Print
         # everything, under each env knob whose name is in the module's string table.
         if: ${{ !cancelled() }}
+        continue-on-error: true
         run: |
-          set -uo pipefail
+          set +e -u -o pipefail
           bundle=bundle/windows-gnu-x64-release
           exe="$bundle/mimalloc-test-stress-dynamic.exe"
           run_one() {
@@ -461,6 +462,15 @@ jobs:
           uv run ci/pe_imports.py "$exe"
           echo "--- mimalloc.dll ---"
           uv run ci/pe_imports.py "$bundle/mimalloc.dll"
+          echo "::endgroup::"
+
+          echo "::group::DIAG redirect-first: does it redirect now?"
+          ( export PATH="$PWD/$bundle:$PATH" MIMALLOC_VERBOSE=1; timeout 60 "$exe" 2 5 1 ) > rf0.raw 2>&1
+          rc=$?
+          if [ "$rc" -ne 0 ]; then echo "VERDICT=FAILED(rc=$rc)";
+          elif grep -q 'malloc is redirected' rf0.raw; then echo "VERDICT=REDIRECTED";
+          else echo "VERDICT=NOT-REDIRECTED"; fi
+          grep -v "^option '" rf0.raw
           echo "::endgroup::"
 
           echo "::group::DIAG redirect-first plain run"
--- a/.github/workflows/windows-bundles.yml
+++ b/.github/workflows/windows-bundles.yml
@@ -452,8 +452,9 @@ jobs:
         # layout crashed at load in run 33614876511 with no output at all; this step gets
         # the faulting instruction and a stack instead of an exit code.
         if: ${{ !cancelled() }}
+        continue-on-error: true
         run: |
-          set -uo pipefail
+          set +e -u -o pipefail
           bundle=bundle/windows-gnu-x64-redirect-first
           exe="$bundle/mimalloc-test-stress-dynamic.exe"
           echo "::group::DIAG redirect-first import tables"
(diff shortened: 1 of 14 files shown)


zackees:
Opus review: APPROVE after fixes. Behavioural proof from the cross-built bundles (release, shared, debug-full): `REDIRECT_BEHAVIOURAL=1` (`mi_is_in_heap_region(malloc(1))`); MSYS2 native msvcrt exe: `REDIRECT_BEHAVIOURAL=0` with `REDIRECT_FLAG=1`, proving the earlier control was false. Root cause (two bugs in series): the redirect module refuses to patch an already-initialised ucrtbase (`LDRP_PROCESS_ATTACH_CALLED`), so `mimalloc.dll` must import `mimalloc-redirect.dll` before its CRT api-sets (`MI_MINGW_REDIRECT_FIRST`, ld sorts `.idata$2` by spelled path); then GCC emulated TLS recursed through the live override (`__emutls_get_address` → malloc), fixed by Win32 TLS slots behind one `MI_WIN_TLS_SLOTS` macro via `_mi_prim_tls_key_*`. Vendored amalgamation regenerated; `-L./<rel>` safe for embedded consumers; docs and #277 corrected.

Issue #272, opened 2026-09-01
Bun parity P7: background scavenger, page hole purging, and mi_on_thread_idle


zackees:
**Depends on:** #266, #267, #271 (and #270 for the fork handler hook). **Blocks:** #274.

Sub-issue of #264 (Phase 7). **Prerequisites: Phases 1, 2, 6 merged.** The largest phase; may be split into 7a (scavenger + `mi_on_thread_idle`) and 7b (hole purging) as two PRs, in that order, if the agent judges it necessary — but 7a's `mi_on_thread_idle` is incomplete without 7b (see contract below), so both must land before Phase 9.

### Why
`mi_on_thread_idle` is **called by Bun today**: `src/jsc/bindings/BunJSCEventLoop.cpp` (event-loop idle, non-Windows), `src/threading/ThreadPool.rs` (worker futex wait timeout), `packages/bun-usockets/src/eventing/epoll_kqueue.c` (rate-limited before parking). Missing symbol = link failure. Its contract (Bun `mimalloc_sys.rs:26-29`): "collects this thread's pending frees, discards the free blocks inside its still-used pages, and hands the arena purge to the scavenger. Safe on any thread; a no-op on a thread that never allocated." Two of the three clauses are the scavenger and hole purging, both **default-on** in Bun (`mi_option_scavenger=1`, `mi_option_purge_holes=1`), built to fix oven-sh/bun#39844 (heap peaks 1.4–2.6× Node). A `mi_collect(false)` shim links but regresses that bug.

### Bun's material
- API: `include/mimalloc.h:203-205` — `mi_on_thread_idle(void)`, `bool mi_on_thread_idle_start(void)` (about to block: hand theaps to scavenger; false = nothing handed off), `mi_on_thread_idle_end(void)` (take them back). Implementations `src/theap.c:311,325,352`. Only `mi_on_thread_idle` is confirmed called by Bun; `_start`/`_end` are used by their park handoff (`test-park-handoff.c`, 560 lines). Implement all three.
[...]

PR #289, opened 2026-09-02
feat: pthread_atfork fork-safety handlers with lock-order contract

### LINKED ISSUE #270: Bun parity P5: pthread_atfork fork-safety handlers with a POSIX repro
**Depends on:** #266. **Blocks:** #272 (child-side scavenger reset), #274.

Sub-issue of #264 (Phase 5). **Prerequisite: Phase 1 merged** (handlers live in `src/subproc.c`, which exists only after the bump). POSIX only; no-op on Windows.

### Why
`grep -rn pthread_atfork include src` returns nothing. `fork()` from a multithreaded process while another thread holds `heaps_lock` / `theaps_lock` / `arena_reserve_lock` / `mi_thread_locals_lock` leaves the child with a permanently-held lock; the child's first `malloc` hangs. Bun runs a multithreaded runtime that spawns processes and registers handlers; `test/test-fork-user-heap.c` (250 lines) is their regression. MIMALLOC_FORKS.md rated this 3 "needs a POSIX repro" — this phase produces the repro first.

### Bun's implementation
- `src/init.c:15` `#include <pthread.h>`; `:618` — `pthread_atfork(&_mi_process_fork_prepare, &_mi_process_fork_parent, &_mi_process_fork_child);` registered **once per process** in process init (Bun commit by Dylan Conway fixed an earlier once-per-`mi_heap_new` registration that exhausted glibc's atfork table and made BoringSSL abort — do not register per heap).
- `src/subproc.c:333` — `static _Atomic(uintptr_t) mi_fork_depth;` nested-fork guard; `:378-389` `_mi_process_fork_prepare` (take locks in a fixed order), `:391-405` `_mi_process_fork_parent` (release in reverse), `:407-420` `_mi_process_fork_child` (reset locks, reset depth, reset scavenger via `_mi_scavenger_forked_child`, re-init thread-local bookkeeping).
- `src/prim/osx/alloc-override-zone.c:387` — the malloc-zone `force_lock`/`force_unlock`/`reinit_lock` callbacks call the same three functions (currently no-ops in ours: `mi__malloc_fork_prepare/parent/child`).
- Documented lock-ordering rule worth keeping verbatim: **locks that can be held across an allocation must be taken before `arena_reserve_lock`.**
- Bun's handlers also touch the scavenger and park state; those parts belong to Phase 7. Port only the lock skeleton and `threadlocal.c` handling here; leave a clearly marked `// Phase 7: scavenger` hook point.

### Steps
1. **Repro first** (Linux + macOS): `test/test-fork-locks.c` — thread A loops `mi_heap_new`/`mi_heap_delete` (holds `heaps_lock`) and large allocations (hits `arena_reserve_lock`); main thread `fork()`s 200 times, child does `mi_malloc(64)` + `mi_free` + `_exit(0)`, parent `waitpid` with a 5 s alarm. Without handlers this should hang or fail some fraction of runs; record the observed failure rate in the PR. Add the test with a timeout property in CMake.
2. Enumerate every process-wide lock in our tree (`grep -rn "mi_lock_t\|_mi_lock_init\|mi_lock_acquire" src include`) and write the order into a comment block at the top of `src/subproc.c`. Include our fork-only locks: `prof_lock` (`src/profile.c`), the DHAT lock (`src/dhat.c`), memory-events callback state (`src/memory-events.c`), `mi_thread_locals_lock` (`src/threadlocal.c`).
3. Implement the three handlers in `src/subproc.c` with provenance; register once in `src/init.c` under `#if !defined(_WIN32) && !defined(__wasi__)` (check upstream's existing platform guards in `init.c`).
4. Wire the macOS zone callbacks (`src/prim/osx/alloc-override-zone.c`) to the handlers; keep nested-call safety via `mi_fork_depth` since the zone callbacks and `pthread_atfork` both fire.
5. Child-side: profiler, DHAT and memory-events must either be reset or safely continue. Decide per component and write it down: recommended — profiler continues (records are process memory, copy-on-write is fine, `prof_lock` reset), DHAT continues, memory-events callbacks continue. `mi_prof_dump` in the child must work (add to test).
6. Import Bun's `test/test-fork-user-heap.c` with provenance (drop the scavenger-specific parts behind `#if 0 // Phase 7` if they don't compile).
7. Rust: no bindings needed; regenerate amalgamation.

### Acceptance
- `test-fork-locks` and `test-fork-user-heap` green on ubuntu and macos in `c-unit`, 0 hangs in 200 forks × 10 CI runs.
- Lock order documented in `src/subproc.c`; `MI_DEBUG` build asserts the order in `_mi_process_fork_prepare`.
- `MIMALLOC_FORKS.md` `pthread_atfork` row → IMPORTED.

Refs #264

### Getting Bun's source
```sh
```

zackees:
## Summary

- Implements #270 (Bun parity P5): `pthread_atfork` fork-safety handlers with a
  documented lock-order contract, plus a POSIX repro test landed first per the
  issue's Step 1.
- `src/subproc.c` gains `_mi_process_fork_prepare/parent/child`, a 13-step lock
  order documented in a comment block at the top of the file and self-checked in
  `MI_DEBUG` builds (`mi_fork_lock_order_assert`).
- Registered once in `src/init.c`'s `mi_process_init_once`, POSIX only
  (`#if !defined(_WIN32) && !defined(__wasi__)`) -- never per heap (see "what the
  issue got wrong" below for why that distinction matters).
- Wired into macOS's `alloc-override-zone.c`: the introspection `force_lock`/
  `force_unlock`/`reinit_lock` callbacks and the `_malloc_fork_*` DYLD interpose
  both call the same handlers, made idempotent by a nesting guard.
- Child-side policy for the profiler/DHAT/memory-events: **continue**. Their state
  is ordinary process memory (profiler/DHAT records come from the raw-OS-layer
  arena per rule 4, never the hooked path) that survives fork() by plain
  copy-on-write; only each subsystem's own lock needs resetting. `mi_prof_dump`/
  `mi_dhat_dump` are exercised in the child by `test-fork-locks.c`.
- `test/test-fork-user-heap.c` imported from `oven-sh/mimalloc@942b8342` (MIT),
  with the two cases needing the not-yet-landed scavenger/tld registry disabled
  as `#if 0 // Phase 7` (they don't compile against this tree).
- `MIMALLOC_FORKS.md`'s `pthread_atfork` row -> IMPORTED.

## Repro (issue Step 1)

`test/test-fork-locks.c`: a churn thread holds `heaps_lock`/`arena_reserve_lock`
(`mi_heap_new`/`mi_heap_delete` + a large allocation in a loop) while the main
thread `fork()`s 200 times; each child does `mi_malloc(64)`/`mi_free`/`_exit(0)`,
watchdog-guarded (5s `alarm()` + `waitpid`).

At the committed test's own scale (1 churn thread, 200 serialized forks) this
reproduced **0/200 failures** on the pre-#270 tree, across 10+ runs (~2,000 forks
total) on this machine -- the race window is narrow, matching Bun's own
"Case A is probabilistic" caveat for their equivalent test. Since the issue asks
for an observed failure rate, I went further with an ad hoc, heavier
multi-threaded stress harness (not part of the committed suite: several forking
threads racing several heap-churn threads). That **did** reproduce the hazard
concretely on the pre-#270 tree: multiple single-threaded child processes left
**permanently hung**, confirmed via `/proc/<pid>/wchan` = `futex_do_wait` with
`Threads: 1`, never progressing (a single-threaded process blocked forever on a
futex is the unambiguous signature of an inherited, permanently-locked mutex --
nothing else produces it). After the handlers, the same heavier harness produced
zero hangs.

## Verification

- `test-fork-locks` / `test-fork-user-heap`: green x20 in Release, green x20 in
  Debug+`MI_DEBUG_FULL`+`MI_PPROF=ON`, no hangs.
- Full `ctest` suite (36 tests, Debug+FULL+`MI_PPROF=ON`; 24 tests,
  `MI_PPROF=OFF`): 100% pass in both.
- `uv run ci/verify_local.py --keep-going`:

  | config | result |
  |---|---|
  | release | PASS |
  | off | PASS |
  | debug-full | PASS |
  | guarded | PASS |
  | shared | PASS |
  | bundle | PASS |
  | memory-gate | PASS |
  | diag | PASS |
  | rust | PASS |
  | lint | PASS |
  | asan | FAIL (environmental -- see below) |

- `asan` is environmentally red in this sandbox: `sanitizer/asan_interface.h` is
  not installed, so CMake falls back to `MI_TRACK_ASAN=OFF` and `verify_local`
  correctly flags that as a failed gate. Confirmed this is unrelated to this PR --
  even `mimalloc-test-api` (no fork, untouched by this change) segfaults with zero
  ASan output under the same manual build.
- A separate manual `gcc -fsanitize=address -fno-omit-frame-pointer` build (not
  through `verify_local`, `MI_DEBUG=2`) was attempted per the task instructions:
  it compiles clean, but running *any* test under it segfaults before printing an
  ASan report (confirmed on `test-api` too, not fork-specific) -- consistent with
  the same environmental gap, not a regression from this PR.
- `pyright` (`uv run --with 'pyright[nodejs]==1.1.411' pyright`): 0 errors.
- `cargo run -p xtask -- amalgamate-c && amalgamate-h && check`: clean; `cargo
  test -p mimalloc-pprof`: all green (regenerated in a separate, rust-only
  commit per rule 2).

## Known limitations / what the issue got wrong

1. **`mi_tld_t::theaps_lock` is not covered.** Bun's ported skeleton quiesces a
   per-subprocess `tld` registry (`sp->tlds`/`tlds_lock`) that walks every live
   thread's `tld->theaps_lock`. That registry is scavenger/park infrastructure
   that does not exist in this tree yet (lands with #264 item 7 / #272). Per the
   issue's own instruction to "port only the lock skeleton... leave a clearly
   marked `// Phase 7: scavenger` hook point," this gap is documented at every
   affected site in `subproc.c` and in `MIMALLOC_FORKS.md`, rather than
   approximated with an ad hoc walk (a heap's `theaps` list can reach the same
   `tld` through more than one theap -- one thread, several heaps -- and locking
   a non-recursive mutex twice inside the handler itself would turn a rare hang
   into a guaranteed one).
2. **The ported `mi_fork_depth` design (Bun's, as specified by the issue) assumes
   fork() calls are serialized across threads. glibc does not do this** --
   verified empirically with a minimal `pthread_atfork` probe: concurrent
   `fork()` calls from different threads interleave their prepare/parent
   handlers essentially every time under load. The single-process-wide-atomic
   nesting counter, exactly as ported, is only correct for same-thread nesting
   (the macOS zone callback + `pthread_atfork` firing together for *one*
   `fork()`); under genuinely concurrent multi-threaded `fork()` it produced a
   real `internal_lock_release_by_non_owner` (an ABA race across overlapping
   fork "generations", caught by the P1 reentrancy checker -- not a false
   positive). Fixed here: `mi_fork_depth` is now a per-thread (`mi_decl_thread`)
   reentrancy count, and a real mutex (`mi_fork_serialize_lock`) gives true
   cross-thread exclusion around one fork "generation". This is not called out
   in the issue at all; I found it during evidence-gathering for the repro
   numbers above and fixed it rather than leaving a known-broken design in a
   safety-critical path.
3. Also found (and independently confirmed unrelated, not fixed here, flagged
   for the user): a **pre-existing** assertion failure
   (`mi_theap_is_valid`, `theap.c:71`) under sustained multi-threaded
   `mi_heap_new`/`mi_heap_delete` churn -- reproduces with **zero** forking
   involved (a single churn thread run alone for ~90s trips it), on both the
   pre-#270 and post-#270 tree. Out of scope for this PR; worth its own issue.
4. macOS CI is cross-built: the arm64 lane compiles only, and x64 execution runs
   under `dockur/macos` pending a golden image (per `.github/workflows/cross.yml`).
   So the zone/DYLD-interpose wiring in `alloc-override-zone.c` is
   **compile-verified only** in this PR -- not exercised by a running macOS test.

## Merge gates checklist (CLAUDE.md rule 3)

- [x] `c-unit` on ubuntu with `MI_PPROF=ON` -- verified locally (verify_local +
      manual full ctest), CI pending.
- [x] `MI_PPROF=OFF` job -- verified locally (manual build + ctest), CI pending.
- [x] `rust-native` -- verified locally (`cargo test -p mimalloc-pprof`), CI
      pending.
- [ ] windows-MSVC / windows-MinGW / macos -- not run locally (no such toolchain
      here); relying on CI. macOS execution is cross-built/compile-only per above.

Closes #270

---

## Update: fixed after Opus review (both blockers reproduced and fixed)

- **Blocker 1 (compile break on macOS/Windows, all non-Linux CI red)**:
  `_mi_process_is_initialized` was declared only inside `prim-tls.h`'s
  `#if MI_TLS_MODEL_LOCAL` branch; macOS (pthreads TLS model) and Windows don't
  take that branch. Fixed: unconditional `extern` in `internal.h` (dropped
  `page.c`'s now-redundant local extern); the whole fork-handler definition
  block in `subproc.c` is now guarded with the same
  `#if !defined(_WIN32) && !defined(__wasi__)` `init.c` already uses to call
  it, so it's not dead code there either (rule 6).
- **Blocker 2 (deterministic AB-BA deadlock under `MIMALLOC_PROF=1`)**:
  reproduced exactly as described (3/3 hangs, Release, `MI_PPROF=ON`).
  Root cause confirmed: Bun's ported rule assumed hook locks
  (`prof_lock`/`dhat_lock`/`memevt_cb_lock`) are never taken from inside the
  allocation path itself, which is false in this tree (`_mi_prof_on_alloc`
  etc. are called from `alloc.c`/`page.c` while a heap's `arena_pages_lock`
  can still be held up the stack). Fixed: hook locks now sit innermost in the
  documented order, alongside `out_buf_lock`; also corrected the
  `heap_main` vs. non-main-heap `arena_pages_lock` ordering per the review's
  note on `mi_heap_ensure_arena_pages`'s real nesting. Verified: the exact
  `MIMALLOC_PROF=1` Release repro now passes 3/3 (and 20x); added `MIMALLOC_DHAT=1`
  / `MIMALLOC_MEMORY_EVENTS=1` ctest variants alongside it so a future
  reordering regresses loudly in CI.
- **Q1 (`mi_decl_thread` violated the codebase's own #266 rule)**: fixed --
  `mi_fork_depth`/`mi_fork_lock_level` are no longer `mi_decl_thread`.
  Replaced with a shared `mi_fork_owner` (`_Atomic mi_threadid_t`) + plain
  `int` depth, both protected by a real mutex (`mi_fork_serialize_lock`),
  needing no TLS at all. This also fixes the asymmetric-guard-latch gap
  (`parent`/`child` now check ownership symmetrically with `prepare` instead
  of blindly decrementing).
- **Q3**: comment now says "modern glibc (>= 2.34)".
- **Secondary**: `dhat_once`/`memevt_once` now reset in the child (new
  `_mi_atomic_once_fork_child_reset` helper) unless already resolved; added
  `MI_DEBUG`-only test hooks (`_mi_test_hold_heaps_lock` etc.) so
  `test-fork-locks.c` has a deterministic repro alongside the probabilistic
  one (which measured 0/200 on `main`, confirmed near-zero discriminating
  power on this machine).
- Documented the residual, pre-existing `mi_prof_visit`/`prof_lock`
  inversion (visitor callback holds the lock, so an allocating visitor
  inverts the alloc-hook nesting) as a "visitor must not allocate" contract
  in `profile.h` -- out of scope to fix in a fork-safety phase.

Verified: `test-fork-locks` (base + all 3 env variants) and
`test-fork-user-heap` green 20x in Debug+FULL and 20x in Release, including
the exact review repro (`MIMALLOC_PROF=1`, Release) now 3/3 clean where it
was 3/3 hung. Full `ctest` green in debug-full (39 tests), release (32),
and `MI_PPROF=OFF` (27). `uv run ci/verify_local.py --keep-going`: 9/11 pass
(`asan` still environmentally red -- confirmed via `test-api`, unrelated to
this PR). Rust amalgamation regenerated in a separate commit;
`cargo test -p mimalloc-pprof` green.

Commits: `73f22a43` (original), `da934391` (original rust), `ee890758`
(review fixes), `77a72e4d` (review-fix rust amalgamation).


zackees:
## Update: review round 2 (F1 + F2)

Commits: `cf15bb2b` (C core), `7d8e203c` (rust amalgamation, separate per rule 2), `c0f622c3` (stress-test scoping), `0a10992b` (merge of `main`).

> The merge was needed to get CI to run at all: after #286 and #290 landed, this PR went `CONFLICTING` (`src/static.c`'s include list — #286's `heap-dump.c` next to #270's `fork.c`) and GitHub silently schedules **no** `pull_request` workflow runs for a PR whose merge commit cannot be computed. Conflict resolved, amalgamation regenerated over the merged tree, all suites re-run below. It also picks up #290's `timed_allocation_audit` fix, which was the standing `rust-native` win-gnu failure.

### F1 — the order is now derived, not asserted

You were right: the outer steps were never re-derived. `_mi_meta_zalloc` holds `theap_meta_lock` across a full `mi_theap_zalloc` on `theap_meta`, whose heap **is** `heap_main` (`subproc.c:339`), so an ordinary allocation slow path runs inside that lock. I re-read every `mi_lock*` site in `src/` and what each critical section can call, and wrote the result out as an edge table at the top of the new `src/fork.c`. Summary (`X -> Y` = some real path holds X while acquiring Y):

| lock | can acquire while held | evidence |
|---|---|---|
| `mi_subprocs_lock` | `heaps_lock`, `heap->theaps_lock` | `_mi_subproc_prof_sync_force_slow`, `_mi_subprocs_unsafe_destroy_all` |
| `sp->heaps_lock` | `theaps_lock`, `mi_thread_locals_lock`, `arena_pages_lock`, `os_abandoned_pages_lock`, `theap_meta_lock` | `mi_subproc_unsafe_destroy` -> `_mi_thread_locals_done` / `_mi_heap_force_destroy` -> `mi_heap_free` (heap.c:203) |
| `heap->theaps_lock` | `theap_meta_lock` | `mi_heap_free_theaps` (heap.c:174) -> `_mi_theap_decref` -> `_mi_meta_free` (theap.c:363); `tld->theaps_lock` only via **try**-acquire (theap.c:412), not a blocking edge |
| non-main `heap->arena_pages_lock` | `heap_main->arena_pages_lock`, `arena_reserve_lock`, page-map lock, `theap_meta_lock`, `os_abandoned_pages_lock` | `mi_heap_ensure_arena_pages` (arena.c:685) holds it across `mi_arena_pages_alloc` -> `mi_heap_zalloc_aligned(heap_main,…)` (arena.c:1544); `mi_heap_free` frees under it |
| `mi_thread_locals_lock` | `theap_meta_lock` | `_mi_thread_local_create` -> `mi_thread_local_create_expand` -> `_mi_meta_zalloc_aligned` (threadlocal.c:349) |
| **`sp->theap_meta_lock`** | **`heap_main->arena_pages_lock`, `arena_reserve_lock`, page-map lock, `heap_main->os_abandoned_pages_lock`** | `_mi_meta_zalloc` (subproc.c:181) -> `mi_theap_zalloc(theap_meta)` -> `_mi_arenas_page_alloc` -> arena.c:784/534, page-map.c:393 |
| `heap_main->arena_pages_lock` | — (leaf) | for a main heap `mi_heap_ensure_arena_pages` only stores `&arena->pages_main`; `mi_heap_free` skips the loop (`if (!is_main)`) |
| `arena_reserve_lock` | — (leaf) | `mi_arena_reserve` -> `mi_arena_initialize` -> `mi_arenas_add` is raw-OS + atomics |
| `os_abandoned_pages_lock`, page-map lock | — (leaf) | list splice / `_mi_os_zalloc` |
| `prof_lock`/`dhat_lock`/`memevt_cb_lock` | (innermost) | taken by the alloc/free hooks, i.e. under anything above |
| `out_buf_lock` | (last) | reachable from a warning under any lock above |

Derived order — a topological order of that graph, which is **not** one walk per subprocess (a process-global lock sits between two per-subprocess levels, and a heap's three locks do not share a level), so prepare runs one pass per level:

```
1 mi_subprocs_lock            6 per sp: theap_meta_lock          11 prof_lock
2 per sp: heaps_lock          7 per sp: heap_main->arena_pages    12 dhat_lock
3 per sp/heap: theaps_lock    8 per sp: arena_reserve_lock        13 memevt_cb_lock
4 per sp/non-main heap:       9 per sp/heap: os_abandoned_pages   14 out_buf_lock
    arena_pages_lock         10 page_map->lock
5 mi_thread_locals_lock
```

Note `theaps_lock` (3) and non-main `arena_pages_lock` (4) end up on opposite sides of `theap_meta_lock` (6) from each other's old grouping — `theaps_lock` must precede it, main `arena_pages_lock` must follow it.

**`mi_fork_lock_order_assert` now checks something real.** The old assert only verified prepare's own sequence — true by construction. Under `MI_DEBUG>2` there is now a runtime detector: `diagnostic.c`'s reentrancy checker already records each lock's owning thread in `debug_owner`, so `_mi_fork_lock_order_observe` (called from `_mi_lock_debug_after_acquire`) reads off which tracked locks the acquiring thread already holds and records the nesting edge in a global bitmap; every `fork()` asserts that all edges observed so far agree with the documented order. Prepare's own acquires are excluded (they hold every level at once in the documented order, which would make the check tautological). Only process-lifetime locks are tracked (main subproc's 3, process `heap_main`'s 3, the 5 globals) — a non-main heap's locks are freed with the heap and address-keyed tracking of them would read freed memory. The whole F1 cycle is inside the tracked set.

**Detector verified end to end**: deliberately swapping the documented levels of `mi_thread_locals_lock` and `theap_meta_lock` (and prepare's matching acquire order) makes an ordinary `test-fork-locks` run print

```
mimalloc: error: fork lock-order violation: mi_thread_locals_lock (step 6) was held while
acquiring subproc->theap_meta_lock (step 5) -- the order documented in src/fork.c is wrong,
or a new nesting was introduced
```

and abort. Restored: silent, and the full 40-test Debug+FULL suite runs clean with no violation reported — i.e. nothing the suite exercises contradicts the derived order.

**Rule 6**: handlers, order table, checker and test hooks moved to a new `src/fork.c` (POSIX-only; `src/static.c` and CMake include it under the same guard). `subproc.c` keeps two accessors (`_mi_subprocs_head` / `_mi_subprocs_lock`); `init.c` is unchanged apart from the comment pointer.

### F2 — the deterministic hook now discriminates

You were right that it had none: the child re-inits `heaps_lock` unconditionally, so no-op'ing the acquire still passed. The holder thread now **poisons the structure the lock guards** while holding it — it detaches `sp->heaps` (saving the head) and restores it before releasing — and the child reports via the new `_mi_test_heaps_lock_poison_observed()`, exiting 3.

| prepare's `heaps_lock` acquire | `deterministic_hold_repro` |
|---|---|
| no-op'd (counterfactual) | **20/20 children observed the poisoned window** (`FAIL: child N forked inside the poisoned heaps_lock window -- _mi_process_fork_prepare did not acquire subproc->heaps_lock`) |
| restored (this tree) | **0/20**, 0 hangs |

The false claims in the file header (lines 31-36) and in the deterministic-repro design note (142-160) are rewritten; `check_dump_in_child` now asserts `prof_ok`/`dhat_ok` when the corresponding subsystem was actually started in the parent.

### Honest scope of the new stress variant

New ctest variant `test-fork-locks-spawn-prof-env` (`MI_TEST_FORK_SPAWN=1` + `MIMALLOC_PROF=1`) keeps up to 64 threads live while forking 200 times, so every thread start goes through `_mi_meta_zalloc` under `theap_meta_lock`. But it is a **timing** workload, not a proof, and I am not going to claim more than I measured:

* The edges the detector actually observes in that run are `mi_subprocs_lock -> heaps_lock -> theaps_lock` and `mi_thread_locals_lock -> theap_meta_lock`. The `theap_meta_lock -> arena/page-map` edges are real by inspection but **rare at runtime** — a meta allocation only reaches `mi_heap_ensure_arena_pages` / `mi_arena_reserve` / page-map growth on a cold page, and the meta theap recycles freed tld/theap slots. Tried `MIMALLOC_ARENA_RESERVE` down to 32 MiB and 64 live threads; they still did not fire.
* Running the variant against the **old** (inverted) order did not hang here in 200 forks. So F1 was a *latent* inversion, not an immediately reproducible one, and the deterministic evidence for the fix is the detector experiment above, not a hang/no-hang pair. The stress variant is the cheap always-on complement.
* Spawn mode **replaces** the heap-churn thread rather than running alongside it: sustained `mi_heap_new`/`mi_heap_delete` churn concurrent with sustained thread starts trips the pre-existing `mi_theap_is_valid` assertion (theap.c:71) this PR already flagged as item 3 — reconfirmed here 3/5 runs of a standalone churn+spawn program that **never calls fork()**, and 0/6 with the spawner alone. Keeping them in separate variants avoids importing a known unrelated flake into CI.

### Verification

* `test-fork-locks` base + `MIMALLOC_PROF=1` + `MIMALLOC_DHAT=1` + `MIMALLOC_MEMORY_EVENTS=1` + spawn/prof, and `test-fork-user-heap`: **20/20 each, in Debug+`MI_DEBUG_FULL`+`MI_PPROF=ON` and in Release** (240 runs, 0 hangs) — re-run in full after the merge, same result.
* Full `ctest`: 41/41 Debug+FULL+PPROF, 34/34 Release (40/33 before the merge added #286's test), 28/28 `MI_PPROF=OFF`.
* **gcc ASan** (`-DMI_TRACK_ASAN=ON -DMI_DEBUG_FULL=ON -DMI_PPROF=ON`, so `MI_DEBUG=3` + `MI_GUARDED=1`): configures with `MI_TRACK_ASAN=ON` and runs clean — `test-fork-locks` base and spawn/prof, and `test-fork-user-heap`, all pass with no ASan report. This also pins down the earlier "environmentally red" claim: `verify_local`'s asan gate configures with **clang** (`/run/current-system/sw/bin/clang`), whose resource dir on this NixOS box has no `sanitizer/asan_interface.h`, so CMake downgrades to `MI_TRACK_ASAN=OFF` and the gate correctly fails. With the default `cc` (gcc) the header is found and the sanitizer really is enabled. Not a property of this PR either way.
* `cargo run -p xtask -- amalgamate-c && amalgamate-h && check` clean (`fork.c` is inlined under the same platform guard); `cargo test -p mimalloc-pprof` green.
* `uv run ci/verify_local.py --keep-going` (post-merge): release, off, debug-full, guarded, shared, bundle, memory-gate, diag, lint **PASS**; `asan` FAIL for the clang-toolchain reason above; `rust` FAIL on `bench-harness --test rejections::benchmark_rejects_partial_or_extra_child_output` — that one is load-induced flake, not a regression: it passes **5/5** run standalone on this exact tree, it is a pure-Rust harness test that spawns `stress-child` with a 10 s `max_duration_secs` (11 configs compiling in parallel is enough to blow it), it touches no mimalloc C, and CI's `rust-native` is green on this SHA. The pre-merge run of `verify_local` was 10/11 with only `asan` red.
* `fork.c` compiles warning-free under `-Wall -Wextra` in Release, Release+`MI_PPROF=OFF` and Debug+FULL.

CI on `0a10992b`: **8/9 green** — `c-unit` (all 17 jobs: MSVC `windows-latest`, `win-gnu`, ubuntu, shared, guarded, debug-full, `ctest-pprof-off`, memory-gate, diagnostic-gates, bundle-roundtrip, isa-baseline x64+arm), `rust-native`, `asan`, `cross`, `fuzz`, `amalgamation-drift`, `doc-snippets`, `benchmark-sentinel` all **success**. The one red is `macos-bundles`: `No golden macOS disk under cache key 'macos-x64-golden-ventura-v1'` — the #277 image has not been published yet, so that job cannot bootstrap itself; it is red on `main` too and is unrelated to this PR. macOS coverage here is therefore `cross` (arm64 compile) only, as already noted in the PR description.

zackees:
Opus review (three rounds): APPROVE. Final round reproduced both proofs: swapping two documented levels trips the MI_DEBUG>2 lock-order detector (`fork lock-order violation: mi_thread_locals_lock (step 6) was held while acquiring subproc->theap_meta_lock (step 5)`), and removing the `heaps_lock` acquire makes the poisoning fork test fail (`child forked inside the poisoned heaps_lock window`); restored, 18/18 across Release / Debug FULL / gcc ASan × {base, PROF, DHAT, MEMORY_EVENTS, spawn-stress, user-heap}. Nesting edges spot-checked in source (theaps→theap_meta via `_mi_theap_decref`; non-main arena_pages→heap_main via `mi_arena_pages_alloc`; theap_meta→heap_main arena/page-map via `_mi_meta_zalloc`). The glibc ≥2.34 concurrent-fork claim was independently confirmed (prepare depth 4–5 on glibc 2.42), so `mi_fork_serialize_lock` is load-bearing.

Follow-up (non-blocking, tracked on #264): anonymous `mi_atomic_do_once` sites (page-map.c:365, profile.c:889, prim-tls.c:124/165/182, init.c) are neither acquired in prepare nor reset in the child — first-init only, narrow.

Merging; only red is the designed `run-macos-x64-dockur` (golden image pending).

Issue #280, opened 2026-09-02
flaky: benchmark-suite timed_allocation_audit fails intermittently on rust-native test (windows-latest)


zackees:
## Symptom
`benchmark-suite/tests/timed_allocation_audit.rs::ordinary_measured_region_performs_zero_harness_allocations` fails intermittently in `rust-native.yml` job `test (windows-latest)`, including on docs-only commits to main.

## Evidence
main runs (rust-native): 33546799829 (f280d00f, docs), 33545491537 (858eb225, docs), 33542883405 (5a9b6acf, docs) — all failed only `test (windows-latest)`; 33565889265, 33546524567, 33546208983, 33544264825 passed. Also red on PR #278 (workflow-only change) run job 100123740542 and on PR #276 runs 33579151991 / 33573550193.

Failure text (job 100123740542):
```

thread 'ordinary_measured_region_performs_zero_harness_allocations' (6924) panicked at benchmark-suite\tests\timed_allocation_audit.rs:195:5:
assertion `left == right` failed: Rust control-plane allocation entered the measured interval for TinyFixed16
  left: 1
 right: 0
stack backtrace:
```

## Why it matters
A merge gate that fails on docs-only commits gets ignored. The test asserts zero Rust control-plane allocations inside the measured interval; on the shared Windows runner something allocates there nondeterministically (timer/thread wake path, or the harness itself under load).

## Ask
Reproduce under load on Windows, identify the allocation (the audit hook should record a backtrace or at least the size/thread), and either fix the harness allocation or make the audit tolerate the specific benign case with a named allowlist — not a blanket retry.

Refs #170, #277.

zackees:
Duplicate of #236 (and #206); evidence moved there.

PR #290, opened 2026-09-02
fix(benchmark-suite): keep harness allocations out of the measured interval (#236)

### LINKED ISSUE #206: ci: rust-native is red on main - timed_allocation_audit fails on macOS, windows, and win-gnu
`rust-native` has been failing on `main` since #202 merged, and the documented merge gate ("`rust-native` green") therefore cannot currently be satisfied by any PR.

## Evidence

A control branch cut from `main` at `7f3bfef9` with a **single added comment line** and no other change (PR #205, closed) reproduces the failure:

| Job | Control from `main` | Test |
|---|---|---|
| `test (ubuntu-latest)` | success | — |
| `test (macos-latest)` | failure | `latency_sampling_buffers_do_not_allocate_after_warmup` |
| `test (windows-latest)` | failure | same |
| `test-win-gnu` | failure | same |

`gh run view 31759288988`.

## What fails

`rust/benchmark-suite/tests/timed_allocation_audit.rs` asserts that zero harness-side (non-adapter) allocations occur inside the measured interval:

```
assertion `left == right` failed: latency sampling allocated in the measured region for TinyFixed64, control=false
  benchmark-suite/tests/timed_allocation_audit.rs:228
```

The audit uses a process-global `MEASURED: AtomicBool` plus a global-allocator wrapper that counts any allocation on any thread while the flag is set, excluding threads currently inside the adapter via a `thread_local` guard. Worker threads live inside that window, so the assertion is sensitive to anything that allocates on a non-adapter thread during the interval.

Two observations that should guide the fix:

1. **The failing test is not stable across runs.** On one branch the same file reported `ordinary_measured_region_performs_zero_harness_allocations` (line 185, `TinyFixed16`) instead. Both tests take `TEST_GATE`, so they do not overlap each other; whichever runs first is the one that reports. That points at a systemic leak of allocations into the window rather than a defect specific to the latency path.
2. **It spread platforms without a code change.** macOS and windows have been red since #202; `test-win-gnu` was green on `main` as recently as run 31553123468 (2026-08-12) and is now red on an unmodified tree, so part of this is runner-image dependent.

Worth checking whether the guard itself allocates on first use: the hook reads `INSIDE_ADAPTER.with(Cell::get)`, and on a platform where thread-local registration allocates lazily, the very first access on a worker thread could recurse into the counting allocator *before* the guard is set.

## Why this matters beyond the red X

The audit is the only thing proving the harness does not perturb the measurements it publishes. While it is red on three of four platforms, that guarantee is unverified everywhere except Linux — and Linux is where the published benchmark numbers come from, so the one platform that matters is still covered, but the safety net for the others is gone.

## Related

- Unblocks the merge gate for #204 and any later benchmark PR.
- `cross` is separately red on `main` (`planted_serialized_control_is_rejected` in `bench-harness`), and `benchmark-sentinel`'s Windows job currently fails to compile at all because the soldr v0.9.0 broker cannot spawn its daemon (`daemon route unavailable ... ErrorBackendSpawnFailed`). Those are distinct problems; this issue is only about `timed_allocation_audit`.

### LINKED ISSUE #236: timed_allocation_audit flaps on ubuntu: a stray allocation enters the measured region

zackees:
## Summary

`rust/benchmark-suite/tests/timed_allocation_audit.rs` runs two `#[test]` fns that
share process-global state (`MEASURED`, `HARNESS_ALLOCATIONS`, the measured-region
hook) and assert that zero non-adapter allocations happen inside a measured
interval. Under libtest's default harness, `#[test]` fns run concurrently, each on
its own OS thread, and libtest initializes per-thread output capture the instant
that thread starts -- an allocation with no ordering relative to this file's
`MEASURED` gate. That stray allocation from the *sibling* test's thread could land
inside the other test's measured interval and get misattributed to the code under
audit, producing:

```
assertion `left == right` failed: Rust control-plane allocation entered the
measured interval for TinyFixed16
```

**Evidence**: reproduced locally on Linux under CPU-pinned contention
(`taskset -c 0,1`, `--test-threads=4`): ~10% failure rate (3/30), with the exact
panic message and line (`timed_allocation_audit.rs:195`) seen in CI (#236, #206,
duplicate #280). `--test-threads=1` under the same contention: 0/60. `TEST_GATE`
only serializes the two tests' own work against each other -- it does nothing
about libtest's per-thread setup on the sibling test's thread, which is why the
mutex didn't help.

## Fix

- `rust/benchmark-suite/Cargo.toml`: declare the test with `harness = false`.
- `rust/benchmark-suite/tests/timed_allocation_audit.rs`: replace the two
  `#[test]` fns with plain functions called sequentially from a hand-written
  `fn main()`. With no libtest harness, there is no sibling test thread in the
  process during either measured interval, so the interference is removed at the
  source rather than retried, tolerated, or allowlisted. `TEST_GATE` and its
  poison-recovery handling are removed as no longer needed (nothing left to
  serialize against).
- The audit hook now also records the size of the first offending allocation (no
  thread id or backtrace capture -- that would itself allocate inside the hook)
  so a future regression names something about the culprit instead of a bare
  `left == 0 / right == 1`.

## Test plan

- [x] `cargo test -p benchmark-suite --test timed_allocation_audit` x50 (via
      `cargo test`, checking exit code): 0 failures
- [x] Direct binary run under the exact `taskset -c 0,1` contention scenario that
      previously reproduced the flake x60: 0 failures
- [x] `cargo test -p benchmark-suite` (full crate): all tests pass
- [x] `rustfmt --check` on the changed file: clean
- [x] `cargo clippy` on the changed test target: no new warnings (pre-existing
      warnings in unrelated `src/` files are untouched by this change)
- [ ] CI: `rust-native` green on ubuntu/windows-latest/win-gnu

Closes #236
Closes #206

### DIFF
--- a/rust/benchmark-suite/Cargo.toml
+++ b/rust/benchmark-suite/Cargo.toml
@@ -46,6 +46,18 @@ path = "src/bin/benchmark-scaling-run.rs"
 name = "benchmark-scaling-validate"
 path = "src/bin/benchmark-scaling-validate.rs"
 
+[[test]]
+name = "timed_allocation_audit"
+path = "tests/timed_allocation_audit.rs"
+# Custom `fn main` (no libtest harness): libtest runs `#[test]` fns
+# concurrently on their own OS threads by default, and it initializes
+# per-thread output capture on each test's thread the moment that thread
+# starts -- an allocation with no ordering relative to this file's
+# `MEASURED` gate. That stray allocation from the *sibling* test's thread
+# landed inside this test's measured interval and was misattributed to
+# the code under audit. See #236 / #206.
+harness = false
+
 [dependencies]
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["float_roundtrip"] }
--- a/rust/benchmark-suite/tests/timed_allocation_audit.rs
+++ b/rust/benchmark-suite/tests/timed_allocation_audit.rs
@@ -3,7 +3,7 @@ use std::cell::Cell;
 use std::collections::HashMap;
 use std::ptr::NonNull;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
-use std::sync::{Mutex, PoisonError};
+use std::sync::Mutex;
 
 use benchmark_suite::execution::{
     execute_cell, execute_latency_cell, set_measured_region_hook, AllocatorAdapter,
@@ -12,17 +12,11 @@ use benchmark_suite::scenarios::{cards, CardId, ScenarioCell, Topology};
 
 static MEASURED: AtomicBool = AtomicBool::new(false);
 static HARNESS_ALLOCATIONS: AtomicU64 = AtomicU64::new(0);
-// Serializes the two audits: they share `MEASURED`, `HARNESS_ALLOCATIONS`, and the
-// process-wide measured-region hook, so they cannot overlap.
-//
-// Acquired with `unwrap_or_else(PoisonError::into_inner)`, never `unwrap()`. These
-// tests assert, and an assertion panic while holding the guard poisons the mutex --
-// so a plain `unwrap()` made the *other* test panic on `PoisonError` at its lock
-// line. One real failure was reported as two, the second pointing at a lock
-// acquisition that has nothing to do with allocations. Recovering is correct rather
-// than a workaround: the guard provides mutual exclusion only, and a `()` payload
-// cannot be left in an inconsistent state. See #236.
-static TEST_GATE: Mutex<()> = Mutex::new(());
+// Size in bytes of the first allocation that tripped `HARNESS_ALLOCATIONS` in
+// the current measured interval, `u64::MAX` when none has. Recorded so a
+// failure names *something* about the culprit without the audit hook itself
+// allocating (a `Backtrace` capture here would recurse into this allocator).
+static FIRST_OFFENDING_SIZE: AtomicU64 = AtomicU64::new(u64::MAX);
 
 thread_local! {
     static INSIDE_ADAPTER: Cell<bool> = const { Cell::new(false) };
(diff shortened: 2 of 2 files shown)


zackees:
Reviewed: rust-only, test-harness change. Root cause (libtest sibling test thread's per-thread capture setup allocating inside the other test's measured interval) reproduced locally at ~10% under pinned contention and eliminated at 0/60; matches the CI panic line exactly. Windows rust jobs green on this PR. Merging.

Issue #66, opened 2026-07-31
Phase C: upstream pin decision + hunk classification


zackees:
Sub-issue of #61 (Phase C). **Blocks any third-party patch import.**

We are pinned to `upstream/dev3` `579f8c0e`; upstream is at v3.4.3+; Bun's fork is rebased on the newer base. Importing across that gap without classification is how forks become unrebasable.

### C.0 — classify (blocks C.2)
For every candidate hunk, decide `{already-upstream, upstream-fixed-differently, fork-only}` with the upstream SHA cited for the first two.
- **already-upstream** → importing is churn, or a conflicting second fix
- **upstream-fixed-differently** → importing creates permanent divergence that conflicts on every rebase and gets rejected at upstreaming time
- **fork-only** → import it

Candidates: Bun's `mi_theap_merge_stats` guard against concurrent `_mi_theap_free`; arena bits + page map cleared on alloc-fresh failure paths.

### C.1 — decide the pin
Bump to the dev3 tip and re-verify every hook patch, **or** record a dated decision to stay pinned with the re-verification cost accepted. "Moving bought nothing last time" is a sunk-cost argument; the cost recurs and grows.

### C.2 — import only `fork-only` hunks, each with a test that fails without it
Note Bun's tests depend on `mi_debug_fail_os_commit_after`, which we do not have — so importing the test corpus is a **core-file change** under rules 2 and 6.

Provenance required: `// imported from oven-sh/mimalloc @ <sha>, MIT`, a row in `MIMALLOC_FORKS.md`, never squashed into an unrelated commit. Our goal is upstreaming to Microsoft; they will not take code whose origin we cannot state.

### Acceptance
Classification table published; pin decision recorded and dated.

Refs #61

zackees:
## C.1 — pin decision: **bump to the dev3 tip.** Measured, not estimated.

The issue warned that "moving bought nothing last time" is a sunk-cost argument and that the re-verification cost recurs and grows. Agreed — so I measured the cost instead of arguing about it.

**Gap:** our pin `579f8c0e` → `upstream/dev3` tip `d077de56` = **92 commits**.

### Upstream churn in the files we hook

| File | Upstream churn in the gap |
|---|---|
| `src/alloc.c` | 91 lines |
| `src/free.c` | 88 lines |
| `src/init.c` | 57 |
| `src/options.c` | 50 |
| `include/mimalloc/internal.h` | 46 |
| `src/page.c` | 27 |
[...]

PR #291, opened 2026-09-02
fix: heap delete/destroy teardown protocol; import Bun's heap test corpus and commit-fail injection

### LINKED ISSUE #271: Bun parity P6: heap delete/destroy teardown protocol + import Bun's heap test corpus and commit-fail injection
**Depends on:** #266. **Blocks:** #272, #274.

Sub-issue of #264 (Phase 6). **Prerequisite: Phase 1 merged.** Blocks Phase 7 (park handoff builds on this protocol).

### Why
Bun's `MimallocArena` (`src/bun_alloc/MimallocArena.rs:102,201`) does `mi_heap_new` per parse/request and `mi_heap_destroy` on `Drop`, from many threads, at high rate, with cross-thread frees into heaps that are being torn down. That pattern produced six Bun PRs (oven-sh/mimalloc #22–#28) and Bun commit `8286bfb6` "heap delete/destroy: one teardown protocol instead of per-reader fixes" (`src/theap.c` +346, `src/heap.c` +61, `page->heap` made non-const). We have none of it and no test that exercises the pattern. Bun will not rebase onto a fork that cannot run their heap regression suite.

### Bun's material
- Commits (in order): `de802667` bootstrap library in `mi_heap_new`/`mi_subproc_new`; `36e8fc33` improve concurrent heap_delete and thread termination; `ec238987` `mi_heap_delete`: detach theaps before pages move, free after; `8286bfb6` the protocol; `ff96441a` free an abandoned page without touching its heap after it is unpublished; `d9a4ae0b` idle_start park fields (Phase 7); `49006f33` make tests pass on every CI config.
- Tests (all in `test/`): `test-heap-teardown.c` (571), `test-heap-mt.c` (359), `test-heap-churn.c` (461), `test-heap-aba.c` (129), `test-heap-delete-race.c` (90), `test-theap-sentinel.c` (227), `test-park-handoff.c` (560 — **Phase 7**, skip here), `test-stress-heaps.c`, `test-commit-fail.c` (98).
- Fault-injection hook the tests need: `include/mimalloc/internal.h:359` `extern mi_decl_export volatile long mi_debug_fail_os_commit_after;` implemented in `src/os.c`, `MI_DEBUG`-gated: after N commits, `_mi_os_commit` fails. `test-commit-fail.c:58` sets it to 1000.
- Also from the same delta: arena bits and page-map cleared on alloc-fresh failure paths (`src/arena.c`), and `mi_theap_merge_stats` guarded against concurrent `_mi_theap_free` (MIMALLOC_FORKS.md pass-3, rated 4).

### Steps (tests first, then fix)
1. Port `mi_debug_fail_os_commit_after` (`src/os.c` + `internal.h`) with provenance. This is a core-file change under rules 2/6 (noted in #66); keep it to the guarded lines.
2. Import `test-commit-fail.c`, `test-heap-teardown.c`, `test-heap-mt.c`, `test-heap-churn.c`, `test-heap-aba.c`, `test-heap-delete-race.c`, `test-theap-sentinel.c`, `test-stress-heaps.c` verbatim with provenance headers; register in `CMakeLists.txt` with the timeouts Bun uses (`test/CMakeLists.txt` at 942b8342 has them). Comment out with `#if 0 // Phase 7` any use of `mi_on_thread_idle*`, `_mi_scavenger_*`, `mi_purge_holes_*`.
3. **Run them and record which fail** on all four platforms (expected: several, under `MI_DEBUG_FULL` especially). Paste the matrix in the PR. If everything passes, our tree may already be safe on some paths — say so, and still import the tests.
4. Port the protocol from `8286bfb6` (+ `ec238987`, `ff96441a`, `36e8fc33`, `de802667`) into `src/theap.c`, `src/heap.c`, `src/free.c`, `src/page.c`, `include/mimalloc/types.h`. Read each commit's message; they document the invariants. `page->heap` becomes non-const. Keep our profiler hook (`page->metadata` / `has_metadata`) consistent: a page that is unpublished from its heap must not be visited by `mi_prof_visit` or DHAT with a dangling heap pointer — add an assertion in `MI_DEBUG`.
5. Classify each hunk `{already-upstream at 6def7be9, upstream-fixed-differently, fork-only}` per #66 before porting; only `fork-only` hunks are imported. (Some of the earlier race fixes may have been upstreamed between `bcee5a88` and `6def7be9`.)
6. Port the `arena.c` alloc-fresh failure cleanup and the `mi_theap_merge_stats` guard, each with a test that fails without it (`test-commit-fail` covers the first).
7. Run `test-profile-race` and `test-subproc-lifecycle` (ours) plus the imported set 20× locally under `MI_DEBUG_FULL` and ASan (`asan.yml`) before opening the PR.
8. Trailing Rust-only commit: amalgamation; also run the Rust stress harness (`rust/stress-harness`) since `mi_heap_*` is used there.

### Acceptance
- All imported tests green on ubuntu/MSVC/MinGW/macos, `MI_PPROF` ON and OFF, `ctest-debug-full`, and `asan`.
- Hunk classification table in the PR (per #66).
- `MIMALLOC_FORKS.md`: "v3 theap/arena race fixes" and "adversarial test corpus" rows → IMPORTED.

Refs #264, #66, #78

### Getting Bun's source
```sh
```

zackees:
## Summary

Implements #271 (Bun parity P6): imports Bun's heap delete/destroy regression
corpus and `mi_debug_fail_os_commit_after` fault-injection hook, then ports
the heap delete/destroy teardown protocol itself, fixing the exact
`mi_theap_is_valid` assertion reproduced from PR #289 (P5)'s evidence
(heavy multi-threaded `mi_heap_new`/`mi_heap_delete` churn, no forking).
Along the way, two use-after-free classes surfaced once the protocol
actually worked well enough to reach them in practice, and are fixed here
too (one is Bun's own `ff96441a` fix applied where this tree also needs it;
the other is specific to this fork's own always-on hooks and vanilla
`mi_stat_free`, not present in Bun's tree at all).

Ten commits, in order:

1. **test:** import `test-commit-fail.c`, `test-heap-teardown.c`,
   `test-heap-mt.c`, `test-heap-churn.c`, `test-heap-aba.c`,
   `test-heap-delete-race.c` (all `oven-sh/mimalloc @ 942b8342`, MIT), plus
   the `mi_debug_fail_os_commit_after` hook (`src/os.c`). Records the "before"
   failure matrix.
2. **fix:** `_mi_heap_detach_theaps` must clear `theap->heap`, not
   `theap->tld` (Bun `ec238987`) + a missing `MI_THREADID_DETACHED`
   exception in `theap.c`'s own validity assertions (present everywhere else
   in the tree, never ported here) — fixes the reproduced ABA/assertion bug.
3. **feat:** the four-step claim protocol itself (Bun `8286bfb6` + `ff96441a`,
   adapted for the absence of `pthread_atfork` and the scavenger).
4. **fix:** two use-after-free classes the now-working protocol exposed.
5. **chore(rust):** amalgamation regen, separate commit per rule 2.
6. **docs:** MIMALLOC_FORKS.md.
7. **merge:** `origin/main`, bringing in P5 (#289, `src/fork.c`
   `pthread_atfork` handlers) and P4 (#286).
8. **chore(rust):** amalgamation regen after the merge.
9. **fix:** restore a fork-aware branch in `mi_heap_visit_page_claim` that
   step 3 had deliberately dropped pending P5 — merging P5 in exposed the
   gap it left (see "What the issue got wrong" below).
10. **chore(rust):** amalgamation regen after the fork-aware fix.

## Hunk classification (per #66; base = `6def7be9`, tip = `942b8342`)

| Commit | Author | Ancestor of `6def7be9`? | Disposition |
|---|---|---|---|
| `de802667` (bootstrap in `mi_heap_new`) | zackees | **yes** | already-upstream, not ported |
| `36e8fc33` (`_mi_heap_detach_theaps` skeleton) | daanx (Microsoft) | **yes** | already-upstream, not ported |
| `0e150b5c` (`claim_pages` / `mi_heap_visit_page_claim` skeleton `8286bfb6` builds on) | Jarred Sumner | no | fork-only; ported (this tree had no equivalent claim loop at all before this PR) |
| `ec238987` (detach clears `theap->heap`) | robobun | no | fork-only; ported (commit 2) |
| `8286bfb6` (four-step protocol) | Jarred Sumner | no | fork-only; ported (commit 3), `_mi_process_is_forked_child` branch initially dropped (marked, "re-add when #270/PR #289 merges") — restored in commit 9 after the merge exposed why it was needed |
| `ff96441a` (test-heap-teardown; `page->heap` UAF fix) | Jarred Sumner | no | fork-only; test import (commit 1) + the UAF fix (commit 4). Its `test-purge-holes.c`/CI-matrix hunks not applicable (no hole-purging in this tree) |
| `49006f33` (CI matrix; `mi_theap_is_valid` `MI_THREADID_DETACHED` exception refinement) | Jarred Sumner | no | fork-only; only the `theap.c` assertion-exception hunk applies here, and only its *pre*-refinement (base `MI_THREADID_DETACHED`) form — the `!= _mi_thread_id()` generalization exists to tolerate Bun's scavenger sweeping a parked thread's theaps, which does not exist in this tree (#272) |
| `d9a4ae0b` (park fields) | Jarred Sumner | no | Phase 7 (#272), not ported |
| arena bit/page-map cleanup on alloc-fresh failure | (upstream, pre-`6def7be9`) | **yes** | already-upstream: this tree's `mi_arenas_page_alloc_fresh` already sets the `arena_pages->pages` bit only *after* a successful commit; `test-commit-fail.c` passes cleanly as regression coverage |
| `mi_theap_merge_stats` vs `_mi_theap_free` guard | (historical fork note) | n/a | not applicable: `_mi_theap_free` no longer exists at this pin (confirmed via `git grep`); `_mi_theap_merge_stats` is byte-identical to Bun's `942b8342` version and already guarded by `mi_theap_is_initialized` |
| `_mi_process_is_forked_child` flag (`subproc.c` in Bun) + its uses in `mi_heap_visit_page_claim` (`arena.c`) and `mi_heap_release_pages` (`heap.c`) | robobun/Jarred Sumner | no | fork-only; ported in commit 9, flag relocated to `src/fork.c` (this tree's fork-handling home, not `subproc.c`) since P5 (#289) landed independently of this issue and without an equivalent persistent flag — see "What the issue got wrong" |

## Not imported

- **`test-park-handoff.c`**, and `test-heap-teardown.c`'s `parked` case
  (`#if 0 // Phase 7`): need `mi_on_thread_idle_start/_end` + the background
  scavenger (#272).
- **`test-theap-sentinel.c`**: white-box-reproduces a `(mi_theap_t*)1`
  sentinel-reservation bug in Bun's `mi_heap_init_theap` (`src/heap.c`).
  This tree's `mi_heap_init_theap` never writes a sentinel to the
  thread-local slot — it writes the real theap pointer exactly once, after
  `_mi_theap_create` succeeds — so there is no equivalent bug, and the test
  needs internal symbols (`_mi_theap_default_safe`) this fork's
  differently-shaped theap internals don't have.

## Before / after failure matrix

Debug, `MI_DEBUG_FULL=ON`, `MI_PPROF=ON`, local, single run unless noted:

| Test | Before (commit 1 baseline) | After (final) |
|---|---|---|
| `test-commit-fail` | PASS (regression coverage; see classification table) | PASS |
| `test-heap-teardown` | TIMEOUT (`pin` hangs, hook unwired) | PASS |
| `test-heap-mt` | FAIL — `_mi_arenas_page_unabandon: mi_page_is_owned(page)` | PASS |
| `test-heap-churn` | FAIL — `mi_theap_is_valid: heap_theap==NULL \|\| heap_theap==theap` | PASS |
| `test-heap-aba` | FAIL — SIGSEGV | PASS |
| `test-heap-delete-race` | FAIL — same `mi_theap_is_valid` assertion | PASS |
| standalone repro (8 threads × 90s `mi_heap_new`/`mi_heap_delete`, no forking) | abort in ~5s | PASS ×2 |

Two further, narrower failures surfaced only under repeated parallel runs
(not single-shot) after the protocol landed, both fixed in the "two
use-after-free classes" commit:

| Test | Before that commit | After |
|---|---|---|
| `test-heap-teardown` (Release, `os-pages`) | SIGSEGV ~1/8 runs | 0/56+ across Debug/Release/ASan |
| `test-heap-mt` (Debug, `heap-free-during-delete-abandoned`) | SIGSEGV up to 5/24 runs | 0/56+ across Debug/Release/ASan |

A third, narrower regression surfaced only after merging `origin/main`
(commit 7), which brought in P5's `pthread_atfork` handlers (`src/fork.c`,
#289) and its `test-fork-user-heap.c`. This was a genuine regression against
a previously-green test, not a pre-existing issue — confirmed via a
side-by-side `git worktree` build of `origin/main` alone (3/3 clean, 0/200
child failures per run):

| Test | On `origin/main` alone | After merge, before commit 9 | After commit 9 |
|---|---|---|---|
| `test-fork-user-heap` (`case_a`) | PASS (0/200 child failures ×3) | intermittent FAIL (~1-2/200 child failures per run) — `mi_page_is_valid_init: page->used + free_count == page->capacity`, downstream of this PR's `mi_heap_visit_page_claim` force-seizing a page from a fork-torn theap | PASS (0/4000 child failures across 20 direct runs + ctest runs) |

Root cause: commit 3 had deliberately dropped Bun's `_mi_process_is_forked_child`
branch in `mi_heap_visit_page_claim`, noting "add it back when #270/PR #289
lands" — but #289 landed a handler-only `pthread_atfork` design with no
persistent "we are a forked child" flag, so there was nothing to restore the
branch against until commit 9 added one (see "What the issue got wrong").

## Verification

- Full `ctest`: **40/40** (Debug, `MI_DEBUG_FULL=ON`, `MI_PPROF=ON`) and
  **27/27** (Release, `MI_PPROF=OFF`), both 100%, no regressions in the
  existing suite (`test-subproc-lifecycle`, `test-stress-heaps`,
  `test-stress-subprocs`, `test-profile-race`, etc).
- Imported corpus run 20× locally (issue's instruction), both configs:
  `test-commit-fail`, `test-heap-teardown`, `test-heap-mt`, `test-heap-churn`,
  `test-heap-aba`, `test-heap-delete-race` — **20/20 clean** in Debug+FULL,
  **20/20 clean** in Release.
- **Real ASan** (not the environmentally-degraded fallback this sandbox
  usually falls back to — `clang`/`clang++` from an available alternate
  toolchain, `~/.clang-tool-chain`, has `sanitizer/asan_interface.h` and
  `MI_TRACK_ASAN` actually compiles in): full `ctest` **46/46** (post-merge
  suite, includes `test-fork-user-heap`), plus 4 extra direct runs of
  `test-fork-user-heap` — **4/4 clean** (800 fork-cycles), no ASan reports.
- After commit 9 (the fork-aware claim fix): full `ctest` **47/47** ×3
  consecutive clean runs (Debug, `MI_DEBUG_FULL=ON`, `MI_PPROF=ON`);
  `test-fork-user-heap` alone 20/20 direct runs clean (4000 fork-cycles);
  the imported P6 corpus (`test-commit-fail`, `test-heap-teardown`,
  `test-heap-mt`, `test-heap-churn`, `test-heap-aba`,
  `test-heap-delete-race`) plus `test-fork-user-heap`, 20× each — **140/140
  clean**.
- `cargo run -p xtask -- amalgamate-c && amalgamate-h`: clean, regenerated
  twice more (once for the merge, once for commit 9's fix).
  `cargo test --workspace` and `cargo test -p stress-harness`: all green,
  including the previously-flaky `bench-harness::rejections::
  benchmark_rejects_partial_or_extra_child_output` (fixed by #290, now
  merged in) and `allocator_scenario::
  concurrent_alloc_free_with_profiler_snapshots`.
- `uv run ci/verify_local.py --keep-going` (post-merge, post-commit-9 tree):

  | Job | Result | Notes |
  |---|---|---|
  | release | PASS | |
  | off | PASS | `MI_PPROF=OFF` gate |
  | debug-full | PASS | `MI_DEBUG_FULL=ON`, `MI_PPROF=ON` |
  | guarded | PASS | |
  | shared | PASS | |
  | bundle | PASS | |
  | memory-gate | PASS | |
  | diag | PASS | |
  | lint | PASS | |
  | rust | FAIL in the full parallel run, **PASS in isolation** (`--only rust`, 116.9s) and standalone `cargo test --workspace`/`-p bench-harness --test rejections` (5/5 reruns) — CPU-contention flake from verify_local's 5-way-parallel job scheduling in this sandbox, not a code issue: `rust/bench-harness` has zero commits on this branch (`git log origin/main..HEAD -- rust/bench-harness` is empty) |
  | asan | FAIL (environmental) | verify_local hardcodes `CC=clang CXX=clang++` off `PATH`, and this sandbox's system clang lacks `sanitizer/asan_interface.h` so `MI_TRACK_ASAN` doesn't survive configure -- matches PR #289's own documented experience. Worked around for this PR's own verification by pointing `CC`/`CXX` at an available alternate clang toolchain that does have the header: real ASan, 46/46 (post-merge) then 47/47-equivalent (post commit-9) `ctest`, plus 4/4 extra direct `test-fork-user-heap` runs, no ASan reports |

## What the issue got wrong / additional findings

1. **File list for Step 4 omitted `src/arena.c`**, where the actual teardown
   protocol lives (~200 lines of `8286bfb6`) — the claim/pin loop, the
   `mi_bitmap_clear_once_set` synchronization with the freer, and the
   OS-abandoned-page list claim.
2. **Two use-after-free bugs were found that Bun's ported diff does not
   fix**, because they're not in Bun's tree at all: this fork's own
   always-on `_mi_memevt_on_free`/`_mi_prof_on_alloc` hooks (rule 6) and
   vanilla `mi_stat_free` read `page->heap`/`mi_page_subproc(page)` from the
   cross-thread free path (`mi_free_block_mt`), which is safe for the *page*
   struct (the block being freed keeps it alive) but not for the *heap* it
   points to, once a concurrent `mi_heap_delete` can actually reach the end
   of its walk quickly (which it could not, reliably, before this PR).
   Fixed via `_mi_meta_is_meta_page_safe` (derives the subproc from the
   page's *arena*, which outlives every heap in it, instead of the page's
   heap) and switching `mi_stat_free`'s unrelated `theap_meta`-fallback use
   of `subproc` to `_mi_subproc()` (this thread's own, always safe).
3. **Bun's own CMakeLists (`942b8342`) registers this test corpus with no
   explicit ctest `TIMEOUT`** (the issue implied there were timeouts to
   carry over) — bounded them here (300s) instead, since a still-open race
   or an intentionally-unwired debug hook should fail the run, not hang it
   forever.
4. **`test-theap-sentinel.c`** targets a bug this tree's differently-shaped
   `mi_heap_init_theap` does not have (see "Not imported" above) — not a
   fork-only hunk to port, no equivalent bug to regress-test.
5. **The issue's premise that `_mi_process_is_forked_child` would exist by
   the time this lands (via #270/PR #289) turned out to be only half true.**
   #289 landed real `pthread_atfork` prepare/parent/child handlers
   (`src/fork.c`) that reset locks and subsystem state, but not a
   persistent "we are a forked child" flag — its own scope was fork-safety
   of the lock graph, not the heap-delete page-ownership question this
   flag exists for. Discovered only after merging `origin/main`, when
   P5's own `test-fork-user-heap.c` started failing intermittently against
   this PR's teardown protocol (see the third failure-matrix table above).
   Fixed by porting Bun's flag (`_mi_process_is_forked_child`), homed in
   `src/fork.c` (this tree's fork-handling file) rather than Bun's
   `subproc.c`, set once in `_mi_process_fork_child` and never cleared.

## Merge gates checklist (CLAUDE.md rule 3)

- [x] `c-unit` on ubuntu with `MI_PPROF=ON` — green in CI.
- [x] `MI_PPROF=OFF` job (`ctest-pprof-off`) — green in CI.
- [x] `rust-native` — green in CI.
- [x] windows-MSVC and windows-MinGW (win-gnu) — both green in CI (see review
      round below; this required real fixes, not just a rebase).

## Review round (Opus, on the first push)

CI never ran on the first push (blocked on a vendored-amalgamation conflict
against `origin/main`); once that was resolved, CI surfaced three problems
review alone couldn't have caught without a Windows toolchain:

- **B1** (real bug): `_mi_process_is_forked_child` was defined inside
  `src/fork.c`'s POSIX-only `#if !defined(_WIN32) && !defined(__wasi__)`
  block, but `arena.c`/`heap.c` read it unconditionally on every platform —
  undefined symbol on MSVC/MinGW. Fixed by moving the definition to
  `src/subproc.c` (compiled unconditionally), keeping only the setter in
  fork.c's POSIX block.
- **M3**: documented (not fixed) that the flag is sticky — set once, never
  cleared, matching Bun. Opened #293 (references #272) as the tracked
  follow-up.
- **M4**: `test-heap-teardown.c`'s `pin` case gated its extern declaration
  on `!defined(NDEBUG)` while the real symbol (arena.c) is gated on
  `MI_DEBUG > 0` — the two diverge in this tree's CMake. Fixed by switching
  to `MI_DEBUG > 0` + an `#error` (matching `test-commit-fail.c`), and
  splitting `test-heap-teardown`'s CMake registration into its own
  `if(MI_DEBUG)` block.
- **L5/L7/L9**: documented the one deliberate behavioral deviation from Bun
  in MIMALLOC_FORKS.md, fixed its issue-number reference, and dropped
  `_mi_theap_can_touch`'s pointless `const`-then-cast-away parameter.

CI then surfaced a problem none of the above predicted: **`error C1189: "C
atomic support is not enabled"`** from MSVC's `vcruntime_c11_stdatomic.h`,
in three imported test files (`test-heap-aba.c`, `test-heap-delete-race.c`,
`test-heap-teardown.c`) that `#include <stdatomic.h>` unconditionally on
their C branch — MSVC rejects that without `/std:c11
/experimental:c11atomics` (not set here). Fixed per file, following this
tree's own established precedent (`test-heap-mt.c` / `test-stress.c`'s
"Portable threading / atomics" block, and `include/mimalloc/atomic.h`'s own
`MI_HAS_C11_ATOMICS` detection): small `InterlockedExchange*`-based
load/store wrappers on the MSVC branch, the existing C11/C++ path
untouched elsewhere. Fix commits: c1f08746 (heap-aba, heap-delete-race),
65df22aa (heap-teardown). Verified clean under gcc (C) and g++ (C++17,
`-x c++`) locally, and confirmed green across every Windows CI job
(`build`/`test`/`ctest`/`ctest-debug-full`/`ctest-shared`/`ab`/
`memory-gate` × MSVC and win-gnu) after the fix.

Final CI: all jobs green except `run-macos-x64 (dockurr/macos on Linux)`,
which fails for an environmental reason unrelated to this PR ("No golden
macOS disk under cache key" — this sandbox has no locally-built macOS
image to bootstrap that self-hosted runner from).

Closes #271


zackees:
Opus review: APPROVE after fixes. Protocol port verified with RED→GREEN evidence (reverting the unpublish-read fix → SIGSEGV in test-heap-teardown; reverting the fork-specific one-liners → 8/24 SIGSEGV in test-heap-mt; both 0 after). Hunk classification confirmed against 6def7be9. Fixes since: `_mi_process_is_forked_child` defined unconditionally in subproc.c (was inside fork.c's POSIX guard → Windows link break), MI_DEBUG gating aligned for the stall hook, MSVC `<stdatomic.h>` in three imported tests, docs/table items, follow-up #293 for the sticky fork flag. CI 54/55 with only the designed `run-macos-x64-dockur` red. Merging.

Issue #170, opened 2026-08-06
perf: establish staged upstream-parity and profiler optimization program


zackees:
## Context

The fork is entering a performance phase with one non-negotiable constraint: diagnostic and profiling work must not make normal release allocation slower. We need to compare like-for-like builds, find unintended serialization, profile before optimizing, and preserve correctness, portability, memory use, and tail latency while pursuing speedups.

There is already evidence that the compiled-in-but-stopped profiler is not free. The Release microbenchmark recorded in README/PR #154 measured 11.75 ns/allocation with `MI_PPROF=OFF` versus 20.00 ns/allocation with `MI_PPROF=ON` and profiling stopped (about +70%). The current allocation fast path calls `_mi_prof_on_alloc`, which performs an atomic enabled check; active samples later serialize record creation, stack interning, and page metadata updates under the global `prof_lock`. The free path usually checks `page->has_metadata` before taking that lock.

Several confounders must be isolated rather than attributed to pprof:

- The fork's independent memory-events feature is always compiled and adds one relaxed state check to allocate/free/realloc even when runtime-disabled. Therefore fork `MI_PPROF=OFF` is not structurally identical to upstream.
- `MI_PPROF=ON` adds `-fno-omit-frame-pointer` on non-Windows, which can change code generation independently of the hook.
- `mi_page_t` contains profiler metadata fields and `mi_tld_t` contains profiler sampling state even in an `MI_PPROF=OFF` build. Their size, alignment, and cache-line placement must be compared with upstream.
- Upstream already has optional `MI_OPT_ARCH` and `MI_OPT_SIMD` paths, including AVX2 bitmap scans. We must measure those before inventing overlapping SIMD implementations.
- The live `upstream/dev3` ref can move. The parity baseline for this program is the fork overlay's exact pinned base, `bcee5a88`, unless a separate issue-scoped overlay update is reviewed and verified first.

GitHub-hosted runners are suitable for correctness and smoke tests, but not for accepting small timing wins. Performance claims require repeatable measurements on stable machines.

## Proposal

Run this as ordered phases. Do not start a later optimization phase until the preceding phase has produced reproducible evidence and its regressions are fixed. Use one PR per phase; keep Rust harness and C-core changes in separate commits as required by the repository.

### Phase 0 — trustworthy Rust benchmark and profiling harness

[...]

PR #292, opened 2026-09-02
ci: build windows-msvc test bundles on Linux via soldr clang-cl, run them on the single Windows runner (#277 phase D)

zackees:
Closes the Windows half of #277: the MSVC-ABI test binaries are now cross-built on
`ubuntu-latest` through soldr's `blessed-msvc` lane (clang-cl 21.1.5 + lld-link + llvm-lib
against an xwin-materialised MS CRT and Windows SDK) and executed on the **same single
Windows runner** phase C created. `run-windows-gnu` is renamed **`run-windows`** and now
carries both lanes.

**Windows runner jobs per push: 13 → 2.**

## Rule 3: one native `cl` job is retained, and that is now written down

CLAUDE.md rule 3 makes MSVC a priority platform. **A clang-cl binary is not a `cl`
binary** — clang-cl accepts `__attribute__`s cl rejects, and cl's codegen, TLS lowering and
DLL runtime are its own. So `c-unit.yml`'s **`ctest (windows-latest)` (Release) stays
native and stays a hard gate**: not `continue-on-error`, not scheduled for deletion, with a
comment on the job saying it must not become either without the owner amending rule 3.
Rule 3 and `docs/ci-gates.md` both say so now.

Every *other* `windows-latest` row is informational for the ≥10-push window, each with a
dated TODO: `c-unit.yml`'s `ctest-debug-full` / `ctest-shared` / `memory-gate` (windows
rows only — the ubuntu rows stay hard gates), `rust-native.yml`'s `test (windows-latest)`,
`cross.yml`'s `test (x86_64-pc-windows-msvc)`, and `benchmark-sentinel.yml`'s
`benchmark-sentinel (windows-latest)`.

## The override works in the cross-built artifact, and it is gated behaviourally

The owner requirement — a cross-compiled artifact must behave like a native one — is gated
the way phase C gated it. All three shared-library MSVC bundles must print
`REDIRECT_BEHAVIOURAL=1` from `mimalloc-test-redirect-probe` (a pointer from the CRT's own
`malloc` landing in a mimalloc region), a **hard failure**. `mi_is_redirected()` is not the
gate: phase C established that it reports true on binaries whose allocations were never
touched. `run-windows` also builds the native `cl` Release tree on the same commit, so the
two arms' import tables and probes are printed side by side.

## Three corrections to #277's phase D row, all measured

**1. `mimalloc-redirect.dll` is not import #0, and does not need to be.** Both arms produce:

```
ADVAPI32.dll  mimalloc-redirect.dll  KERNEL32.dll  MSVCP140.dll  VCRUNTIME140.dll
api-ms-win-crt-stdio-…  api-ms-win-crt-runtime-…  …
```

`ADVAPI32.dll` is #0 because `${mi_libraries}` is attached to the target before the
redirect import library is. It is harmless: the module bails only when `ucrtbase`'s
`LDRP_PROCESS_ATTACH_CALLED` is already set, and ADVAPI32 is a KnownDLL whose subtree uses
legacy `msvcrt.dll`. The gate is therefore **"redirect before every CRT module"**, not a
fixed index — asserting #0 would fail a layout that works. lld-link needs no
`MI_MINGW_REDIRECT_FIRST` equivalent.

**2. "clang-cl links against the system UCRT — no extra runtime DLLs expected" is wrong.**
`mimalloc.dll` imports `VCRUNTIME140.dll` **and** `MSVCP140.dll` (the latter because
`CMakeLists.txt` ~205-212 detects clang-cl as MSVC-like and turns `MI_USE_CXX` on itself,
exactly as for `cl` — so the workflow does *not* pass it, it asserts the detection still
fires via `/Zc:__cplusplus`). Those are the Visual C++ redistributable, **not part of
Windows**, and soldr's splat has import libraries only so they cannot be shipped. Declared
via a new `bundle_tests.py --allow-msvc-runtime` (opt-in and lane-scoped, so the win-gnu
lane cannot silently acquire a VC++ dependency) and **checked** with `where` on the runner
before anything runs.

**3. `llvm-rc` is not in soldr's exported toolchain, and CMake's MSVC link rule needs one.**
`cmake -E vs_link_exe --rc=… --mt=… --manifests` embeds a default manifest on every link.
soldr's LLVM ships neither `llvm-rc` nor `llvm-mt`. Without an opt-out the lane's outcome
depends on the host: with a stray `llvm-rc` on `PATH`, CMake bakes an absolute host path
into the cache; without one, configure dies inside `CMakeTestCCompiler` with
`RC Pass 1: command "rc /fo …/manifest.rc" failed`. The toolchain passes `/MANIFEST:NO`,
which `cmVSLink` parses and which skips the rc/mt pass entirely. Verified by configuring
with every `llvm-rc` removed from `PATH`.

## The `/MD` pin, and what debug-full really is

The pin is load-bearing and verified — with that one line removed:

```
lld-link: error: could not open 'msvcrtd.lib': No such file or directory
```

The `windows-msvc-x64-debug-full` bundle is configured **`Release` + `MI_DEBUG_FULL`**, not
`Debug` — and that is not only the CRT constraint, it is what reproduces the native job.
`ctest-debug-full (windows-latest)` passes no `-G`, so the Visual Studio generator takes
the build type from `--config` and leaves `CMAKE_BUILD_TYPE` at its `build`-directory
default of Release. The native job's *resolved* configuration is
`MI_DEBUG=3 MI_GUARDED=1 MI_FREE_IS_CHECKED=1 MI_BUILD_RELEASE`, library `mimalloc` — which
this bundle reproduces exactly. Passing `Debug` would be *less* faithful (it drops
`MI_BUILD_RELEASE` and renames the library). The genuine loss is the compile flavour:
`/MD /O2` vs `/MDd /Od /RTC1`. The assertions and expensive invariant checks — what
`MI_DEBUG_FULL` is for — are identical.

*(This also corrects phase C's correction 4: phase B's "debug-full is really Release +
`MI_DEBUG=3`" finding is not macOS-only, it holds for windows-MSVC too, for the different
reason above. windows-GNU remains the one lane that really configures `Debug`.)*

## Coverage

`build-rust (x86_64-pc-windows-msvc)` runs cargo **twice**, because it replaces two native
jobs that differ in scope *and* in profile — collapsing them would quietly change what is
tested:

| build | stands in for | staged / built |
|---|---|---:|
| `cargo test --workspace --no-run` (**debug**) | `rust-native.yml` `test (windows-latest)` | 49 / 53 |
| `cargo test -p mimalloc-pprof --no-run --release` | `cross.yml` `test (x86_64-pc-windows-msvc)` | 19 / 19 |

**68 test binaries execute on the runner** (`ran 68 MSVC test binaries`). Building only the
narrow set would have cut Windows Rust coverage by more than half while the table claimed
parity.

Two pre-existing Rust-side defects surfaced along the way. Neither is a clang-cl, MSVC or
cross-compilation property — **both reproduce natively** — so both are reported on #277
rather than fixed here (rule 7; "changing what any test asserts" is out of scope for that
issue, and a test fix is a `rust/` commit under rule 2). Full write-up:
https://github.com/zackees/mimalloc-pprof/issues/277#issuecomment-5511396609

1. **Four test binaries cannot leave the machine that built them.**
   `env!("CARGO_BIN_EXE_<name>")` is expanded at *compile time* into the builder's absolute
   path, so `bench-harness`'s `planted_control`/`rejections`/`throughput` and `dashboard`'s
   `stress_child` carry a literal `/home/runner/.../stress-child.exe`. Excluded by
   **detecting the property** (grep each image for the builder's target path) rather than a
   name list that would rot; the scanner self-tests on a NUL-containing control first,
   because `grep -F` without `-a` silently matches nothing on binaries — which is exactly
   how the first cut of this scan was a false green. A `mimalloc-pprof` binary landing in
   that set is a hard error.
2. **`stress-harness`'s `timing_contract` tests only pass in a debug build.** They assert
   `outer_started.elapsed().as_millis() >= 100` around a 100 ms sleep; optimised, that
   lands on 99. The same binary built natively for **Linux** fails 3/3 in release at
   `--test-threads=1` and passes 3/3 in debug, on unmodified `main`. `cargo test` is a
   debug build, which is why no native job has ever run it.

Stated rather than gated, in `docs/ci-gates.md`:

- doctests cannot run from `--tests` binaries (Linux-only since phase B, now on Windows too);
- `xtask check` / `cargo publish --dry-run` / `check_crate_package.py` are platform-
  independent and are now skipped on the Windows row per #277's addendum item 5, rather
  than carried into a bundle;
- the PR sentinel moves into `run-windows` as a cross-built `dashboard.exe`, guarded on
  `github.event_name == 'pull_request'` so folding it into a push-triggered workflow does
  not turn a PR-only benchmark into a per-push one, and `continue-on-error` because #171 /
  #170 already label hosted-runner numbers informational. (Incidentally: `dashboard` prints
  its JSON to stdout and writes no `.json` file, which is why `benchmark-sentinel.yml`'s
  `rust/target/release/*.json` upload has always been empty. The new step redirects it.)

## Memory gate

`--arch x64 --compiler soldr-clang-cl-21`, so it cannot borrow `windows-pprof1.json` which
native `cl` recorded on the same runner. That baseline does not exist yet, so the lane
bootstraps yellow (`where` rc 3) and its positive control is skipped with a warning until
`ci/memory-baselines/windows-x64-soldr-clang-cl-21-pprof1.json` is committed from the
artifact — same follow-up as phase B's correction 8 and phase C's correction 7.

## Verification

`run-windows` is **green in 11m34s** on one `windows-latest` VM: 29/29 + 36/36 + 27/27
win-gnu and 28/28 + 35/35 + 26/26 MSVC against `cl`-configured trees ("Every test in the
native side is present in the bundle side"), `REDIRECT_BEHAVIOURAL=1` on all three MSVC
bundles *and* on the native `cl` build, 68 Rust test binaries, both pprof dump checks, both
memory gates and the sentinel.

Spiked locally on Linux before any YAML was written: all four configs configure and build,
`/MD` confirmed on the artifact (`prim.c.obj` `.drectve` requests `msvcrt.lib`, no `*d.lib`),
`file format coff-x86-64`, `.CRT$XLB`/`$XLY`/`$XIB` present, non-empty Thread Storage
Directory, **no `__emutls` anywhere**, and four bundles with zero absolute paths. `ruff`,
`pyright` (0 errors), `pytest ci/tests` (215 passed, 4 new) and
`uv run ci/verify_local.py --keep-going` all green.

Refs #277. Do not merge before CI.

### DIFF
--- a/.github/workflows/benchmark-sentinel.yml
+++ b/.github/workflows/benchmark-sentinel.yml
@@ -27,6 +27,14 @@ jobs:
     name: benchmark-sentinel (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
+    # TODO(#277 phase D, added 2026-09-02): DELETE the windows-latest row after ten pushes
+    # have compared it against the `PR benchmark sentinel` step in windows-bundles.yml's
+    # `run-windows`, which runs a Linux-cross-built `dashboard.exe` on the shared Windows
+    # VM. Informational until then. The numbers were already labelled informational (#171,
+    # #170: hosted runners are not a stable reference host) and no gate reads them, which
+    # is what makes it acceptable for the replacement to measure on a VM that has just run
+    # four ctest suites. The ubuntu-latest row is NOT affected.
+    continue-on-error: ${{ matrix.os == 'windows-latest' }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/c-unit.yml
+++ b/.github/workflows/c-unit.yml
@@ -18,6 +18,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # THE RETAINED NATIVE MSVC GATE (#277 phase D). Every other windows-latest row in this
+  # repository is now informational, replaced by Linux-cross-built bundles that
+  # windows-bundles.yml's `run-windows` executes on one shared VM. This one is not, and
+  # must not become so without the owner amending CLAUDE.md rule 3 ("MSVC and win-gnu are
+  # priority platforms -- both, always"): the bundles are built by clang-cl, and a
+  # clang-cl binary is not a `cl` binary. clang-cl accepts `__attribute__`s that cl
+  # rejects, and cl's codegen, TLS lowering and DLL runtime are its own. This job is the
+  # only thing in CI that compiles this tree with Microsoft's compiler and runs the
+  # result. Do NOT add continue-on-error here.
   ctest:
     strategy:
       fail-fast: false
@@ -113,6 +122,14 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    # TODO(#277 phase D, added 2026-09-02): DELETE the windows-latest row after ten pushes
+    # have compared it against `run-windows` in windows-bundles.yml, which runs the same
+    # test set from a Linux-built bundle on the shared Windows VM. Informational until
+    # then -- it is the control arm for a deliberate toolchain change (native `cl` /MDd
+    # vs cross clang-cl /MD), so do not delete it before that comparison has actually
+    # happened. A permanently-yellow job is worse than an absent one. The ubuntu-latest
+    # row of this matrix is NOT affected and stays a hard gate.
+    continue-on-error: ${{ matrix.os == 'windows-latest' }}
     steps:
       - uses: actions/checkout@v4
       - run: cmake -B build -DMI_PPROF=ON -DMI_DEBUG_FULL=ON && cmake --build build --config Debug && ctest --test-dir build -C Debug --output-on-failure
@@ -229,6 +246,14 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    # TODO(#277 phase D, added 2026-09-02): DELETE the windows-latest row after ten pushes
+    # have compared it against `run-windows` in windows-bundles.yml, which runs the same
+    # test set from a Linux-built bundle on the shared Windows VM. Informational until
+    # then -- it is the control arm for a deliberate toolchain change (native `cl` /MDd
+    # vs cross clang-cl /MD), so do not delete it before that comparison has actually
+    # happened. A permanently-yellow job is worse than an absent one. The ubuntu-latest
+    # row of this matrix is NOT affected and stays a hard gate.
+    continue-on-error: ${{ matrix.os == 'windows-latest' }}
     steps:
       - uses: actions/checkout@v4
       - run: cmake -B build -DMI_PPROF=ON -DMI_BUILD_SHARED=ON -DMI_BUILD_STATIC=OFF -DMI_BUILD_OBJECT=OFF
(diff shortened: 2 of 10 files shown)


zackees:
Opus review: APPROVE. Independently reproduced with soldr: COFF x86-64, TLS callbacks and directory present, zero `__emutls`, `/MD` pin load-bearing (splat has no `msvcrtd.lib`), import order `ADVAPI32 → mimalloc-redirect → KERNEL32 → MSVCP140 → VCRUNTIME140 → api-ms-win-crt-*` on both cross and native arms, `MI_USE_CXX` auto-detect matches native cl, bundle 28 tests / zero absolute paths, `--allow-msvc-runtime` fails closed. CI 58/59, only the designed dockur red. Non-blocking follow-ups filed as issues (Rust bin exclusion should match by package id; excluded bins lose Windows coverage until relocatable; `where` check derived from actual imports; assert `MI_USE_CXX=ON` directly; native-cl behavioural probe observed not gated; 13→2 is post-deletion arithmetic). Merging.

Issue #293, opened 2026-09-02
Clear _mi_process_is_forked_child once a forked child's thread population resyncs with reality


zackees:
## Context

PR #291 (issue #271, Bun parity P6) ports Bun's `_mi_process_is_forked_child`
flag (`src/subproc.c` for the definition, `src/fork.c`'s
`_mi_process_fork_child` for the setter) to make `mi_heap_visit_page_claim`
(`src/arena.c`) and `mi_heap_detach_theaps` (`src/heap.c`) fork-safe: after a
multi-threaded `fork()`, the single-threaded child may still be holding
page-map/arena-pages bits published by a sibling thread that did not survive
the fork, and this flag tells those two call sites to re-derive ownership
from the bitmaps instead of trusting that thread's (possibly torn) theap
bookkeeping.

## The limitation (matches Bun's own tree, not introduced here)

The flag is set once in `_mi_process_fork_child` and **never cleared** — this
was a deliberate scope decision in #291 (reviewer finding M3), not an
oversight, because clearing it correctly needs to know when the child's
thread population has resynced with reality, i.e. exactly the "child spawns
its own new threads after fork()" question this issue (#272) already covers.

While the flag stays set:
- `mi_heap_visit_page_claim` permanently takes the force-seize branch for
  every page claim during heap delete/destroy in the forked child, even long
  after the child has spawned entirely new, healthy threads of its own.
- `mi_heap_detach_theaps` permanently skips `_mi_theap_abandon` and stats-merge
  for any theap not owned by the *calling* thread, on the same permanent
  assumption that such theaps belong to dead pre-fork threads.

Both are memory-safe as-is (this is not a correctness bug), just needlessly
pessimistic once the process is well past the fork: it forgoes the normal
claim/pin protocol's wait-and-retry cooperation with concurrent frees, and it
under-reports abandoned-page stats.

## Ask

Once #272 lands the background scavenger / thread-idle machinery and this
tree has a real notion of the forked child's live thread set, revisit
whether `_mi_process_is_forked_child` can be cleared (or replaced with a
narrower "this specific theap predates the fork and its owner is confirmed
gone" check) once the child's thread population has resynced.

Not to be fixed independently of #272 — see PR #291 review discussion.

zackees:
#272 Phase 7a landed as PR #299. It does **not** clear `_mi_process_is_forked_child` — that
stays deliberate, per this issue and rule 7 — but it does put the hook point in place, so
recording it here.

[...]

Issue #283, opened 2026-09-02
ci(macos): run the x86_64-apple-darwin bundle in a Linux-hosted macOS guest instead of Rosetta or macos-15-intel


zackees:
## Where this fits

#277 phase B already did the hard half: cross-build the Apple bundles, then execute them on **one** macOS runner (`run-macos` in `macos-bundles.yml`) instead of holding a slot per matrix row. This proposes finishing the same idea for the **x86_64** half specifically, by moving its execution off Apple hardware entirely.

Today `x86_64-apple-darwin` is executed one of two ways, and both have a cost:

1. **Rosetta 2 on the ARM runner** — `run-macos` does `softwareupdate --install-rosetta` and runs the x86_64 bundle translated. It works, but it is not Intel: Rosetta is a binary translator with its own memory-ordering behaviour (it enables TSO), so an allocator's x86 atomics are exercised through a translation layer rather than on the hardware they target.
2. **`macos-15-intel`** — `cross.yml` line 303 still schedules a real Intel runner for that target. Real hardware, but a scarce 10x-billed slot, and the phase-B TODO at line 280 already wants those rows gone.

A Linux-hosted macOS x86 guest gives native x86 execution at 1x billing, and returns both Apple slots to the pool.

## Why this is worth it for an allocator specifically

mimalloc-pprof is exactly the kind of code where the arch difference is the point, not an implementation detail: atomics, memory ordering, page/alignment assumptions, and the MSVC-vs-clang ARM64 intrinsics problem that produced the 0.9.3 floor. Running the x86 suite under a translator is the weakest link in that chain — TSO semantics can mask precisely the ordering bug an x86 build would otherwise expose.

## What I verified downstream (kernal-api), so this is not theoretical

Working through the same problem in `zackees/kernal-api` (PR #85 there):

- **Soldr cross-compiles both Apple targets from Linux.** It carries its own toolchain — bundled LLVM plus a managed macOS SDK at `~/.soldr/bin/apple-sdk/14.5/darwin-{x86_64,aarch64}` — no xwin, zig, or osxcross involved. Verified by linking real Mach-O: `cf fa ed fe 07 00 00 01` (MH_MAGIC_64 + CPU_TYPE_X86_64) and `cf fa ed fe 0c 00 00 01` (CPU_TYPE_ARM64).
[...]

PR #299, opened 2026-09-02
feat: background scavenger thread and mi_on_thread_idle (Bun parity P7a)

zackees:
Implements the 7a half of #272 (Bun parity P7): the background scavenger thread,
`mi_on_thread_idle` / `_start` / `_end`, `mi_scavenger_stop`, the `scavenger` option
(default 1), `purge_delay` 1000 → 100 ms, and the fork integration. Hole purging (7b) is
a separate stacked PR.

Ported from `oven-sh/mimalloc@942b8342`, MIT; every ported block carries the provenance
line.

## What lands

| Piece | Where |
|---|---|
| Scavenger thread: futex / `__ulock_*` / `WaitOnAddress` / pthread-condvar fallback, WASI stubs, all-signals-blocked-except-faults, 30 s safety net | `src/scavenger.c` (new) |
| Park protocol (`park_state` RUNNING/PARKED/SWEEPING, `_mi_park_leave`, `_mi_theap_sweep_parked`, `_mi_thread_idle_work`) and the public `mi_on_thread_idle*` / `mi_scavenger_stop` | `src/scavenger.c` |
| `mi_arena_schedule_purge` wakes on the subproc-expire 0→set edge; `_mi_arenas_purge_now`; `try_purge` leaves `purge_expire` at the earliest pending per-arena expire | `src/arena.c` |
| Per-subproc `tlds` registry (`tlds`/`tlds_lock`/`parked_count`/`scavenger_wake`) + tld park fields | `types.h`, `init.c`, `subproc.c` |
| Fork: `_mi_scavenger_forked_child`, `_mi_park_leave` at the top of prepare, `tlds_lock` as a new lock level, per-tld `theaps_lock` re-init in the child | `src/fork.c` |
| Option `scavenger` (default 1, appended **after** `purge_zeroes`, no slot renumbered); `purge_delay` 100 ms (oven-sh/bun#34217) | `src/options.c`, `mimalloc.h` |
| `synchronization` in `mi_libraries` for **every** Windows toolchain | `CMakeLists.txt` |

`mi_on_thread_idle` today delivers clauses 1 and 3 of its documented contract (collect
pending frees, hand the arena purge to the scavenger); clause 2 (discard free blocks
inside still-used pages) arrives with 7b, at the marked spot in `_mi_thread_idle_work`.

## Design decisions the issue asked for

**Profiler vs. purger** — written down where the code is, not assumed.

1. *Records never inside a discarded range.* Records live in the profiler's raw-OS arena
   (rule 4), so a purge cannot decommit one. The real risk is a record still **attached**
   to a page being handed to `mi_arena_schedule_purge`. That cannot happen — a page only
   reaches the arena once `mi_page_all_free`, and every block free unlinks its record
   under `prof_lock` — and `mi_arenas_page_free_ex` now **asserts** it
   (`!page->has_metadata`) under `MI_PPROF` + `MI_DEBUG`.
2. *`mi_prof_visit` / `mi_prof_snapshot_new` vs. a concurrent purge.* Neither follows a
   `mi_page_t*` at all: both walk the profiler's own tables under `prof_lock`. There is no
   page to read after a decommit, so no liveness check is needed — the empirical half is
   new scenario 4 of `test-profile-race.c` (8 workers churning whole-slice 256 KiB blocks
   and parking, against 200 rounds of `mi_prof_visit` + `mi_prof_snapshot_new` that
   dereference every field of every sample), registered a second time with
   `MIMALLOC_PURGE_DELAY=1`.
3. *A parked thread's `mi_tld_t.profiler` is untouched.* `_mi_theap_sweep_parked`
   snapshots all four fields around `_mi_thread_idle_work` and asserts equality
   (`MI_DEBUG`). Stronger companion check: the scavenger asserts it still has **no theap
   of its own** after every sweep — if a hook or stat on the sweep path ever forced
   `_mi_theap_default()`, the scavenger would acquire this fork's per-thread
   profiler/hook state and register itself as a parkable thread. It does not: every hook
   accessor peeks (`_mi_hooks_tld_peek` → NULL, `_peek_or_local` → the caller's stack).
4. *`purge_zeroes` vs. `purge_holes_eager_zero`* is 7b's to unify; 7a touches neither.

**`mi_tld_t::theaps_lock` — re-initialized in the child, never acquired in prepare.** P5
left this as a documented KNOWN GAP because there was no way to enumerate live tlds. The
registry removes that obstacle and then shows the gap must be closed the other way round:
this lock's holder is *required* to be able to outlive a fork (`mi_thread_theaps_done`
holds it across a whole theap teardown), and `test-fork-user-heap`'s case_b forks exactly
into that window. Acquiring it in prepare turns that case's child-side deadlock into a
**parent**-side one. Verified both ways: with the child re-init case_b passes; commenting
the re-init out makes the child hit its 5 s watchdog (`status=0x200`) every run.

**Lazy start is restricted to a main-subproc thread** (Bun has no such condition).
`pthread_create` allocates the new thread's stack and TLS through the *calling* thread's
malloc, so a scavenger started from a sub-subprocess thread has its TLS in arenas that
`mi_subproc_destroy` unmaps long before `mi_process_done` joins it — a reproducible
SIGSEGV in `_dl_deallocate_tls` under `test-stress-subprocs`. The scavenger only ever
sweeps the main subproc anyway.

## Bugs found on the way (all fixed here)

- **`_mi_prim_thread_yield` never yielded.** It was `sleep(0)`, which glibc implements as
  `nanosleep({0,0})` — a syscall that returns without rescheduling. Every spin-then-yield
  back-off in the tree was a busy spin. `sched_yield()`, from 942b8342 (`0b0153b2`).
- **`_mi_arenas_try_purge`'s one-purger guard was a function-local static.** A fork landing
  inside the guarded section (far more likely now that the scavenger spends real time
  there) left the child with a flag no thread would ever clear — a child that never purges
  again. Hoisted, and reset from the child handler.
- **`_mi_thread_done` must leave only the *calling* thread's park** — on Windows FLS
  shutdown the exiting main thread runs it against other threads' tlds.
- `_mi_theap_can_touch` and `theap.c`'s `MI_DEBUG>=3` validity predicates now admit the
  "owner parked, scavenger claimed it" case, as Bun's do.

## Tests

`test/test-park-handoff.c` imported (two adaptations, marked `Phase 7b` inline: the
"did a sweep run" observable is `_mi_test_idle_work_count()` rather than
`discard_calls`, and the `purge_holes_min_interval` pacing cases wait for 7b), registered
twice — scavenger on **and** off, so every case also exercises its "nobody to hand off to"
branch. `test-heap-teardown`'s `parked` case and `test-fork-user-heap`'s cases B and C
re-enabled, with Bun's `mi_debug_stall_in_thread_theaps_done` hook. Acceptance test
`test/test-thread-idle-rss.cpp`: a C++ TU that declares
`extern "C" void mi_on_thread_idle(void) noexcept;` **by hand**, as Bun's consumer code
does, so a mangling/`noexcept` mismatch is a link error.

## Numbers

Acceptance criterion (RSS drop ≥ 50 % within 1 s after a worker frees 256 MiB and idles,
via `mi_process_info`'s `current_rss`):

| run | held | after idle | dropped | elapsed |
|---|---|---|---|---|
| 1 | 296 MiB | 12 MiB | 288 MiB (97 %) | 5 ms |
| 2 | 296 MiB | 12 MiB | 288 MiB (97 %) | 5 ms |
| 3 | 296 MiB | 125 MiB | 175 MiB (59 %) | 0 ms |

`test-fork-user-heap` case C (a forked child must still purge): child RSS **141 MB → 9 MB**.

## Memory regression table

Peak process RSS (`/usr/bin/time -f %M`), **minimum of 5** runs each — noise on a shared
machine only ever adds memory, so the minimum is the estimator closest to the true
high-water mark (the same rule `ci/memory_gate.py` uses).

| workload | `main` (0c3d7f91) | this PR (defaults) | delta |
|---|---:|---:|---:|
| `test-stress` (4 threads, transfer churn) | 292.4 MB | 262.9 MB | **−10.1 %** |
| `test-stress-heaps` | 909.5 MB | 832.5 MB | **−8.5 %** |
| `test-memory-gate` | 59.5 MB | 59.5 MB | 0 |
| `ci/memory_gate.py check` (allocator peak_rss, min of 8) | baseline 22.3 MB | **21.8 MB, PASS** (allowed 25.6) | −2 % |

Option matrix on this branch (`test-stress`, same estimator), to separate the two knobs:

| `scavenger` | `purge_delay` | peak RSS |
|---|---|---:|
| 0 | 1000 (i.e. `main`'s behaviour) | 272.5 MB |
| 0 | 100 | 263.9 MB |
| 1 | 1000 | 290.7 MB |
| 1 | 100 (**default**) | 262.9 MB |

`test-stress` peak is a genuinely noisy signal (the `scavenger=1, delay=1000` cell is
within its run-to-run spread of the `scavenger=0` one); the load-bearing numbers are the
first table and the acceptance test above, where the process is actually idle — which is
the case #272 exists for.

`test-fork-user-heap` case C, the forked-child purge case re-enabled here: child RSS
**141 MB → 9 MB**.

## Latency

PR #154's microbenchmark methodology (single-threaded `mi_malloc`/`mi_free` pairs over 8
small size classes, N=4 M, best of 5 inside each process), **min and mean of 10 processes**,
alternating between the two binaries on the same loaded machine:

| build | min ns/alloc+free | mean |
|---|---:|---:|
| `main` (0c3d7f91) | 24.86 | 25.44 |
| this PR | **24.20** | **24.96** |

Within noise (and marginally in the faster direction). This is *after* the struct-layout
fix in `perf: append the new mi_subproc_t fields at the tail`: appending
`scavenger_wake` / the tld registry mid-struct the way Bun does, which shifts
`mi_subproc_t::stats` (touched on the free path), measured a **consistent** 26.66 / 28.03 —
a real ~2 ns regression for a struct nothing here needed to reorder. Absolute numbers are
higher than #50's 11.75 ns because this machine is running several agents; the comparison
is interleaved, so the ratio is the meaningful part. Posting these on #10 as well.

## `uv run ci/verify_local.py --keep-going`

| config | result |
|---|---|
| release | PASS |
| off (`MI_PPROF=OFF`) | PASS |
| debug-full | PASS |
| guarded | PASS |
| shared | PASS |
| bundle | PASS |
| memory-gate | PASS |
| diag | PASS |
| rust | PASS |
| lint | PASS |
| asan | FAIL — environmental (`MI_TRACK_ASAN did not survive configure`, this machine's clang; fails identically on `main`) |

ASan covered manually instead, with gcc: `CC=gcc -DMI_TRACK_ASAN=ON`, full ctest,
**43/43 pass** including `test-park-handoff`, `test-profile-race-scavenger` and
`test-thread-idle-rss`.

Two genuine flakes surfaced by running the suite *under load* rather than idle, both fixed
in `fix: a forced arena purge must not lose the race with the scavenger`:

- `_mi_arenas_try_purge`'s one-purger guard is non-blocking, so `mi_collect(true)` silently
  purged **nothing** whenever it lost the race with the scavenger — `test-degenerate`'s
  thread-churn RSS backstop failed 4 runs in 40 (5.1 MB → 54.1 MB). 0/40 after the fix, and
  0/20 with `MIMALLOC_SCAVENGER=0` or `MIMALLOC_PURGE_DELAY=1000`, which is what identified it.
- `test-profile.c`'s `during.heap_committed >= before.heap_committed` is no longer a valid
  invariant with a background purger: ~1 run in 30 under load.

## Windows link, both lanes

Cross-built locally from this branch on Linux:

- mingw (soldr gcc 15.3.0, UCRT) resolves
  `$MINGW/x86_64-w64-mingw32/sysroot/usr/lib/libsynchronization.a`; the executables import
  `api-ms-win-core-synch-l1-2-0.dll`, which `ci/bundle_tests.py` accepts as
  Windows-supplied. 35 tests bundled, 1 toolchain DLL (`libgcc_s_seh-1.dll`, as before).
- clang-cl / lld-link (xwin splat 2026-06-22) resolves
  `$XWIN/sdk/lib/um/x86_64/synchronization.lib`. 34 tests bundled,
  `--check-dll-closure --allow-msvc-runtime` clean.

Both lanes assert the *resolved* `Link libraries` list exactly, so those assertions and
their two `docs/ci-gates.md` copies are updated.

**macOS is compile-and-link-verified only.** The `__ulock_wait`/`__ulock_wake` path
resolves against the 14.5 SDK in the aarch64 soldr cross build; x64 execution is gated on
#283's golden image, so no macOS *run* of the scavenger is claimed.

## CI

63 checks: 62 pass, **2 fail** — one of which is mine and is NOT resolved. See the
pinned comment on this PR for the full evidence on `run-windows`; in short, in the
clang-cl `windows-msvc-x64-release` bundle `test-thread-idle-rss` passes every
assertion and then the process never exits (ctest kills it at 120 s). The same test
from the MinGW bundle on the same VM passes in 0.09 s, and the native `cl` gate is
green, so it is specific to the clang-cl / `MI_USE_CXX` / `/MD` lane. Two candidate
mechanisms are named in that comment; both are mine and both are unbounded waits, so
neither should be "fixed" on a guess without a Windows host to reproduce on.
**This PR is not ready to merge until that is resolved.**

Everything else:

| lane | result |
|---|---|
| `ctest` / `ctest-debug-full` / `ctest-shared` / `ctest-guarded` / `memory-gate` (ubuntu + windows-latest native) | pass |
| `ctest-musl` (Alpine) | pass |
| `ctest-win-gnu`, `ctest-shared-win-gnu`, `ctest-debug-full-win-gnu`, `test-win-gnu` | pass |
| `build-windows-msvc` x4, `build-windows-gnu` x4, `build-macos` x6, `bundle-roundtrip` | pass |
| `rust-native` (7 targets), `cross`, `amalgamation-drift`, `asan`, `fuzz`, `diagnostic-gates`, `isa-baseline`, `benchmark-sentinel`, `zero-tracking` | pass |
| `run-macos-x64 (dockurr/macos on Linux)` | **fail — pre-existing.** "No golden macOS disk under cache key `macos-x64-golden-ventura-v1`", i.e. #283's designed-red image gate. Identical failure on `main` (runs 33654891521, 33647332771, 33646494384). |
| `run-windows` | **fail — introduced here, open.** MinGW bundle 35/35; MSVC bundle 33/34, `test-thread-idle-rss` asserts fine then hangs at process exit. See the comment. |

Four CI-only bugs this PR introduced and fixed in-branch; every one was reproduced
locally first, none by re-pushing and hoping:

1. **musl.** Bun's `scavenger.c` includes `<linux/futex.h>` — a kernel uapi header
   Alpine's `build-base` does not carry. Reproduced in an `alpine:3.20` container;
   fixed by defining `FUTEX_WAIT_PRIVATE`/`FUTEX_WAKE_PRIVATE` (fixed kernel ABI
   values) rather than adding a build dependency (rule 5). Verified with the job's
   exact command line: 42/42 ON, 36/36 OFF.
2. **macOS cross.** `enable_language(CXX)` for the acceptance test dies with
   "Could not find install_name_tool" on soldr's Linux-hosted Apple SDK toolchain.
   Skipped there; the test's claim is covered on Linux, native `cl`, MinGW and clang-cl.
3. **Native MSVC.** `error C2382: '_mi_thread_idle_work': redefinition; different
   exception specifications` — declared without `mi_attr_noexcept`, defined with it.
   Fine in C, fatal compiled as C++. This one failure took down `ctest`,
   `ctest-shared`, `ctest-debug-full`, `memory-gate` and `zero-tracking`'s `ab` on
   windows-latest; all five went green together.
4. `_mi_test_idle_work_count` needed `mi_decl_externc` (MI_USE_CXX mangles it) and
   `mi_decl_export` (the shared lib hides it under `-fvisibility=hidden`).

Worth noting for future Windows work: **the two Windows lanes disagree on the
language of `.c` sources.** `MI_USE_CXX` is on for both, but only native `cl`
actually compiled `scavenger.c` as C++ — the clang-cl cross lane accepted the
noexcept mismatch. Native `cl` is the stricter of the two, which is a good argument
for keeping it as the retained gate (`c-unit.yml`'s note already says so for a
different reason).

## Not in this PR

- Hole purging, `mi_purge_holes_stats_get`/`_report`, `test-purge-holes.c`,
  `test-purge-zero.c` → **7b**. Its first, purely additive slice (`_mi_prim_discard`
  on all platforms, `_mi_os_discard`, the `mi_page_t` bitmap fields, the shared
  inline helpers) is pushed on `bun-parity/p7b-holes`, stacked on this branch, with
  no PR yet: what remains is the engine itself, which is a hot-path change and wants
  the same load-testing discipline that found the three bugs above.
- The `purge_zeroes` "unification" #272 asks for turns out not to exist, twice over:
  `purge_holes_eager_zero` is a **pre-discard `memzero`** that makes a mis-scoped
  discard corrupt visibly on macOS's lazy `MADV_FREE_REUSABLE` — a test-sharpening
  knob, not zero tracking — and `mi_option_purge_zeroes` is a **dead option** in this
  tree: `git grep mi_option_purge_zeroes -- src/` is empty, its implementation having
  been dropped by the pin bump (#80 / PR #113). Evidence and impact (including that
  `zero-tracking.yml` A/Bs a knob that changes nothing) filed on
  [#67](https://github.com/zackees/mimalloc-pprof/issues/67#issuecomment-5513181359),
  and on [#272](https://github.com/zackees/mimalloc-pprof/issues/272#issuecomment-5513187950).
- #293 (clearing the sticky `_mi_process_is_forked_child`): deliberately not done here.
  `_mi_scavenger_start_lazy`'s post-fork restart is the first "a new thread exists in this
  child" evidence in the tree and is the natural hook point; commented on #293.

## ctest-guarded fix-up (post-review)

- **VMA-cap fix**: `test-park-handoff`'s three permanent-leak stress functions
  (`test_park_then_exit`, `test_exit_while_swept_with_dyn_tls`,
  `test_exit_while_swept_stress`) now gate their thread/round counts on the actual
  runtime `guarded_sample_rate`, not just the `MI_GUARDED` compile flag (which is on
  for every `MI_DEBUG` build regardless). `ctest-debug-full` keeps the full counts and
  their #272-regression discriminating power; only `ctest-guarded`'s forced
  `MIMALLOC_GUARDED_SAMPLE_RATE=1` pass scales down, from a measured peak of ~304,606
  VMAs (past the CI runner's `vm.max_map_count` of 65,530) to ~33,000.
- **Rule 6 note**: this branch's `src/arena.c` (+144) and `src/init.c` (+93) insertions
  are new logic in upstream files, not just guarded hook call-sites -- justified by the
  purge-scheduling and tld-registry work needing file-scope statics in those files.
  Flagging per rule 6 (new logic belongs in new files where practical); the additions
  are `arena.c`'s purge-scheduling/one-purger-guard plumbing and `init.c`'s per-subproc
  `tlds` registry, both load-bearing for the scavenger design in this PR, not moved
  elsewhere on review.

Refs #272, #264, #10


zackees:
## Open, unresolved: `run-windows` — the MSVC bundle's process does not exit

Not a lane to explain away — this is a real regression from this PR, and it is the one
thing still red that is not red on `main` (`main`'s `run-windows` is green, run
33654891546).

**What happened.** On the shared Windows VM, in the `windows-msvc-x64-release` bundle:

```
30/34 Test #30: test-thread-idle-rss ................... Failed  120.03 sec
      reason: timed out after 120s
      | rss: start 4 MiB, live 261 MiB (held 256 MiB)
      | rss after idle: 127 MiB (dropped 133 MiB of 256 MiB held) after 0ms
      | test-thread-idle-rss: ok
```

The test's **assertions passed** — 256 MiB held, 133 MiB (52 %) returned within 0 ms,
final line printed. What hangs is everything after `main` returns: the process never
exits, and ctest kills it at 120 s. The other 33 tests in that bundle pass, including
`test-stress`, `test-stress-subprocs`, `test-heap-mt` and `test-profile-race-scavenger`,
all of which also start the scavenger.

**What makes it specific.** On the *same VM*, from the *same commit*, the MinGW bundle
runs the identical test and passes in 0.09 s (`35/35`), and the native `cl` +
native-ctest gate (`ctest (windows-latest)`) is green. So it is not the scavenger's
`WaitOnAddress` path or the C source: it reproduces only in the clang-cl-cross,
`MI_USE_CXX`, `/MD` bundle.

**Two candidate mechanisms, both introduced by this PR, both unbounded by construction:**

1. `_mi_scavenger_stop`'s `WaitForSingleObject(_mi_scavenger_thread, INFINITE)`
   (`src/scavenger.c`). It runs first in `mi_process_done_once`, i.e. on a Windows
   process-teardown path that can execute under the loader lock — where joining a
   thread is the classic deadlock. An `INFINITE` join from teardown is a latent hang
   regardless of which test exposes it.
2. The forced-purge wait added in `fix: a forced arena purge must not lose the race with
   the scavenger` (`_mi_arenas_try_purge`): `do { guard } while (!ran && force)` with a
   `_mi_prim_thread_yield()` between attempts. `mi_process_done_once` reaches it through
   `mi_theap_collect(theap, true)`. If the one-purger guard is ever left set — by a
   thread that did not run its `mi_atomic_guard` epilogue — this spins forever.

Both want a bound; neither should be bounded on a guess. I have no Windows host to
reproduce on, so rather than push a speculative fix I am recording it here with the
evidence. Reproducing it needs either a Windows machine or a CI run that keeps the
bundle and attaches a debugger to the hung process (`procdump -ma` on
`mimalloc-test-thread-idle-rss.exe` after the 120 s mark would name the frame outright).

**Interim options, in the order I would try them:**

1. Give the Windows join a bound (e.g. `WaitForSingleObject(h, 2000)`) and skip the
   `CloseHandle` on timeout. Teardown must never be able to hang; the process is
   exiting anyway and `_mi_scavenger_running` is already 0 before the wait.
2. Bound the forced-purge retry to a large-but-finite count, keeping the fix for the
   race it was added for while making the loop incapable of hanging.
3. If neither is it, `test-thread-idle-rss` is the only *new* executable in the bundle
   and it is the only one that idles for up to a second without allocating — that
   combination, not the assertions, is what to bisect.

Everything else on this PR is green: 62 checks pass, and `run-macos-x64` fails
identically on `main` (#283's missing golden image).

zackees:
Found while writing Phase 7b (PR #302, stacked on this branch) — reporting rather than fixing, since this branch is yours.

**`_mi_deferred_free` is missing Bun's owner-only guard.** Bun's `src/page.c` at `942b8342` opens it with

```c
void _mi_deferred_free(mi_theap_t* theap, bool force) {
  // owner only: `heartbeat` and `recurse` are plain fields, and the handler is thread-affine
  if (theap->tld != NULL && theap->tld->thread_id != _mi_thread_id()) return;
```

This tree's copy has no such check (`src/page.c:1002`). It matters as of 7a: `_mi_thread_idle_work` calls `mi_theap_collect(theap0)` for a **parked** thread's theap, on the scavenger, and `mi_theap_collect_ex` calls `_mi_deferred_free` first thing. So on the scavenger thread we

1. write `theap->heartbeat++` and `theap->tld->recurse` — plain (non-atomic) fields of a theap owned by another thread, and
2. invoke the embedder's `mi_register_deferred_free` handler from a thread they never created. If that handler allocates, it forces `_mi_theap_default()` into existence on the scavenger, which is exactly what `_mi_theap_sweep_parked`'s `mi_assert_internal(!mi_theap_is_initialized(_mi_theap_default()))` exists to prevent — it would fire in a debug build, and in a release build the scavenger would silently acquire this fork's per-thread profiler/hook state and register itself in `subproc->tlds` as a parkable thread.

Nothing in the current test suite registers a deferred-free handler, so this is latent, not a live red. Phase 7b makes the parked sweep do considerably more work per claim but does not change the exposure.

Suggested fix is Bun's one-liner above, in `_mi_deferred_free`. I have not touched it on `bun-parity/p7-scavenger` or on `bun-parity/p7b-holes`.

zackees:
Second 7a finding from the Phase 7b work (PR #302) — this one is a genuine MSVC correctness bug, not a style nit.

**`_mi_theap_can_touch` reads the 32-bit `park_state` through the 64-bit atomic primitive on MSVC.** From the `windows-msvc` build log on my run:

```
include\mimalloc\internal.h(947,11): warning C4133: 'function': incompatible types
    - from 'uint32_t *' to 'const uintptr_t *'
```

Line 947 is

```c
return (mi_atomic_load_acquire(&theap->tld->park_state) == MI_PARK_SWEEPING);
```

with `_Atomic(uint32_t) park_state;`. This is exactly the hazard `src/profile.c` already documents for `prof_enabled`:

> `size_t` (not bool): MSVC's plain-C atomic wrapper (mimalloc/atomic.h) only implements uintptr_t/int64_t width primitives, so an `_Atomic(bool)` here silently reinterprets this 1-byte flag as an 8-byte load/store (MSVC warning C4133), corrupting adjacent globals and reading garbage.

Same thing here, one field wider: on a 64-bit MSVC build the load reads **8 bytes** starting at `park_state`, i.e. `park_state | (park_reclaim << 32)`. So `_mi_theap_can_touch` returns `false` for a genuinely SWEEPING theap whenever `park_reclaim` is non-zero — which is precisely the window `_mi_park_leave` sets it in. It is an assertion helper, so the symptom is a spurious `mi_assert` failure in MSVC debug builds during a park/leave race, not a crash in release; the read itself is also 4 bytes out of bounds of the field.

The same pattern appears on the other `_Atomic(uint32_t)` park fields (`park_state`, `park_reclaim`, `park_swept`) wherever they go through `mi_atomic_load_*` / `mi_atomic_cas_*`; the CAS sites take `uint32_t*` explicitly so they are fine, but every `mi_atomic_load_acquire`/`_relaxed` on them wants checking. `subproc->scavenger_wake` is also `_Atomic(uint32_t)` and is passed to `WaitOnAddress`, so its width matters too.

Suggested shape (matching `prof_enabled`): make the park fields `_Atomic(size_t)`, or add a 32-bit accessor. Not touched on either branch — reporting only.

zackees:
Follow-up to the C4133 comment above, so the 7a CI being green is not mistaken for evidence against it.

Confirmed at the macro level: `mi_atomic_load_acquire(p)` -> `mi_atomic_load_explicit`, whose MSVC-C definition (`include/mimalloc/atomic.h:250`) is

```c
static inline uintptr_t mi_atomic_load_explicit(_Atomic(uintptr_t) const* p, mi_memory_order mo) {
  #if defined(_M_IX86) || defined(_M_X64)
    return (uintptr_t)MI_MSC_XX(__iso_volatile_load)((volatile const intptr_t*)p);
```

`MI_MSC_XX` expands to `__iso_volatile_load64` on x64, so this is an 8-byte load from a 4-byte `_Atomic(uint32_t)` field — the high half is whatever follows it in `mi_tld_s`, which for `park_state` is `park_reclaim`. The compare is against the full 64-bit value, so `_mi_theap_can_touch` reads `false` for a SWEEPING theap exactly while `park_reclaim` is set.

**Why CI does not catch it:** `test/test-park-handoff.c` is `#ifdef _WIN32` -> `printf("skipped on Windows (uses pthreads/fork)"); return 0;` (line 33). The park protocol therefore has no Windows execution coverage at all right now, so the MSVC lanes being green says nothing about this. Same reason it would not show up in Phase 7b's runs.

zackees:
### Pre-existing race in the parked-thread sweep — reproducible on THIS branch, not introduced by 7b

Third finding, and this one is a live failure rather than a latent one. Phase 7b's CI turned `ctest-debug-full (ubuntu-latest)` red on PR #302; I chased it and it reproduces on `bun-parity/p7-scavenger` at head `2fcac37b` **at a higher rate than on the 7b branch**.

```
mimalloc: assertion failed: at "src/theap.c":49, mi_theap_visit_pages
  assertion: "!include_full || count == total"
```

**Reproduction** (both branches, `-DCMAKE_BUILD_TYPE=Debug -DMI_PPROF=ON -DMI_DEBUG_FULL=ON`, gcc, Linux x64):

```sh
for i in $(seq 120); do taskset -c 0-3 ./mimalloc-test-park-handoff >/dev/null 2>err || grep -m1 assertion err; done
```

The `taskset` matters — unpinned it does not reproduce at all in 30 runs; on 4 CPUs it does. Rates over 120 runs each:

| branch | config | failures |
|---|---|---:|
| ** @ ** | default | **3/120** |
|  @  | default | 2/120 |
|  @  |  | 2/120 |

7b's hole sweep makes the parked-thread walk do more work per park, so it is a slightly better *amplifier*, but the rate is the same order and the base branch's is higher.  is irrelevant, which rules out sweep frequency as the cause.

**What is going on.** It always fires inside  (8 threads park/wake in a loop, plus the forced-collect thread), on the **scavenger**, walking a parked thread's theap:

```
#3  _mi_assert_fail (assertion="!include_full || count == total")
#4  _mi_theap_visit_pages (theap=0x20000346000, fn=mi_theap_page_is_valid, include_full=true)
#5  ... mi_theap_collect_ex  (7a) / mi_theap_purge_holes (7b -- both shapes seen)
#8  _mi_theap_sweep_parked
#9  mi_scavenger_run
```

At the assertion:  (the  snapshot taken at entry),  (pages actually walked), and  by then — so the theap **gained two pages during the walk**, with  and .

Two things I ruled out with instrumented builds:

* ** is not stale at entry.** I recounted the real queue length at every  entry over 30 runs: it always equalled . So pages really are being added mid-walk.
* **It is not another thread pushing.** I probed // for any call where , printing a backtrace. The only hits are the sweeping thread itself ( -> , expected — the scavenger's tid legitimately differs from ). No foreign push was ever caught.

So something adds pages to a theap whose owner is supposed to be quiesced by /, and my probes say it is not a queue push from another thread — which points at the park protocol's quiescence guarantee itself rather than at the queue code. I did not get further than that. A second assertion,  's block-conservation check, also fires on the same workload, which is consistent with the theap being mutated under the walk.

**Why CI is usually green on this branch:** at ~2.5% per run,  passes most of the time. #299's own  is green right now; that is luck, not evidence.

Not touched on either branch — this is yours, and 7b sits on top of it. Happy to hand over the probe patches if useful.

zackees:
### Pre-existing race in the parked-thread sweep — reproducible on THIS branch, not introduced by 7b

(Reposting: my previous comment was mangled by shell expansion — please read this one instead.)

Third finding, and a live failure rather than a latent one. Phase 7b's CI turned `ctest-debug-full (ubuntu-latest)` red on PR #302; I chased it and it reproduces on `bun-parity/p7-scavenger` at head `2fcac37b` **at a higher rate than on the 7b branch**.

```
mimalloc: assertion failed: at "src/theap.c":49, mi_theap_visit_pages
  assertion: "!include_full || count == total"
```

**Reproduction** (both branches, `-DCMAKE_BUILD_TYPE=Debug -DMI_PPROF=ON -DMI_DEBUG_FULL=ON`, gcc, Linux x64):

```sh
for i in $(seq 120); do taskset -c 0-3 ./mimalloc-test-park-handoff >/dev/null 2>err || grep -m1 assertion err; done
```

The `taskset` matters — unpinned it did not reproduce in 30 runs; pinned to 4 CPUs it does. Rates over 120 runs each:

| branch | config | failures |
|---|---|---:|
| `bun-parity/p7-scavenger` @ `2fcac37b` | default | **3/120** |
| `bun-parity/p7b-holes` @ `a1c96855` | default | 2/120 |
| `bun-parity/p7b-holes` @ `a1c96855` | `MIMALLOC_PURGE_HOLES_MIN_INTERVAL=0` | 2/120 |

7b's hole sweep makes the parked-thread walk do more work per park, so it is a slightly better *amplifier*, but the rate is the same order and **the base branch's is higher**. Sweep frequency is not the cause either: forcing a sweep on every single park (`min_interval=0`) does not change the rate.

**What is going on.** It always fires inside `test_park_stress` (8 threads park/wake in a loop, plus the forced-collect thread), on the **scavenger**, walking a parked thread's theap:

```
#3  _mi_assert_fail (assertion="!include_full || count == total")
#4  _mi_theap_visit_pages (theap=0x20000346000, fn=mi_theap_page_is_valid, include_full=true)
#5  ... mi_theap_collect_ex        (7a shape)
    ... mi_theap_purge_holes       (7b shape -- both observed)
#8  _mi_theap_sweep_parked
#9  mi_scavenger_run
```

At the assertion: `total = 21` (the `theap->page_count` snapshot taken at entry), `count = 22` (pages actually walked), and `theap->page_count == 23` by then — the theap **gained two pages during the walk**, with `park_state == MI_PARK_SWEEPING` and `park_reclaim == 0`.

Two things I ruled out with instrumented builds:

* **`page_count` is not stale at entry.** I recounted the real queue length at every `_mi_theap_visit_pages` entry over 30 runs; it always equalled `page_count`. So pages really are being added mid-walk.
* **It is not another thread pushing.** I probed `mi_page_queue_push` / `push_at_end` / `remove` for any call where `theap->tld->park_state == MI_PARK_SWEEPING`, printing a backtrace. The only hits are the sweeping thread itself (`mi_theap_page_purge_holes` -> `_mi_page_free`, expected — the scavenger's tid legitimately differs from `theap->tld->thread_id`). No foreign push was ever caught.

So something adds pages to a theap whose owner is supposed to be quiesced by `MI_PARK_PARKED` / `MI_PARK_SWEEPING`, and the probes say it is not a queue push from another thread — which points at the park protocol's quiescence guarantee rather than at the queue code. I did not get further than that. A second assertion, the block-conservation check in `mi_page_is_valid_init`, also fires on the same workload, consistent with the theap being mutated under the walk.

**Why CI is usually green here:** at ~2.5% per run, `ctest-debug-full` passes most of the time. #299's own `ctest-debug-full (ubuntu-latest)` is green right now — that is luck, not evidence.

Not touched on either branch; 7b sits on top of it. Happy to hand over the probe patches if useful.

zackees:
Two corrections/additions to the race report above.

**1. `2fcac37b` already contains your park-field widening.** I fetched `origin/bun-parity/p7-scavenger` fresh for the comparison, so the 3/120 was measured on the head whose commit message is *"chore(rust): re-run the amalgamation; guard the park-field widths in the vendored C"* — i.e. **after** the `_Atomic(uint32_t)` -> `_Atomic(size_t)` fix, not before it. That makes the finding stronger, not weaker: the race survives the atomic-width fix, so that fix is not the answer here. (The 7b branch's merge base is the earlier `8885a8af`; the 2/120 there is on the pre-widening code, which is why 7b's rate is if anything slightly lower — different base, same phenomenon.)

**2. A hypothesis worth probing that I did not eliminate.** My probes ruled out a *foreign queue push* while `park_state == MI_PARK_SWEEPING`, and the sweeping thread's own `_mi_page_free` only ever *removes*. But `count > total` means pages were **added**. The remaining path is that the owner is not actually quiesced for the whole walk:

`total` is snapshotted at `_mi_theap_visit_pages` entry, and I only observed `park_state == 2` at the *assertion*. In `test_park_stress`'s tight loop the state can plausibly go SWEEPING -> (owner's `_mi_park_leave` CAS) -> RUNNING -> owner allocates (`page_count++`) -> owner re-parks -> scavenger claims again -> SWEEPING, all while one `_mi_theap_visit_pages` call is still in flight — which would produce exactly `count > total` with `park_state == SWEEPING` and `park_reclaim == 0` at the end. My push probe would not catch it either, because at the moment of the push the state is RUNNING, not SWEEPING.

If that is it, the bug is in the handoff protocol rather than anywhere in the walk: something needs to make the scavenger notice that the tld left and re-entered the park mid-walk (a park generation counter would do it), or make `_mi_park_leave` wait for the *walk*, not just for the SWEEPING flag.

Suggested probe: log `park_state` **transitions** on the swept tld with a walk in progress, rather than sampling the state at the assertion. I did not chase it further — handing it over rather than guessing in your code.

zackees:
Opus reviews (four rounds): APPROVE. Verified in the final rounds: word-width park/wake atomics with static asserts (MSVC-C build), `tlds_lock`-held detach, start/stop shutdown edge, owner-only `_mi_deferred_free`, the `test-park-handoff` race root-caused (thread exiting while parked) and fixed on the slow paths with fast-path disassembly unchanged, the Windows exit hang root-caused (ExitProcess kills the scavenger before TLS-detach teardown) and fixed via atexit + bounded join, guarded-lane VMA bound (~33k vs 65k cap). CI: 53 pass, macOS lanes gated off, only red is `memory-gate (ubuntu-latest)` at 27.1 MB with 16.2% spread over its 15% tolerance (#298; main reads 24.5–29 MB on the same baseline). Merging.

Issue #55, opened 2026-07-31
CI: transient setup-soldr EPIPE on macOS silently turns a tagged release into a no-op


zackees:
## Release workflow is not resilient to a transient setup failure, and a flake blocks the publish

The real `v0.9.0` release run ([30645619888](https://github.com/zackees/mimalloc-pprof/actions/runs/30645619888)) failed on `macos-latest` and the publish job was **skipped**, so nothing was published until the job was re-run by hand.

The failure was not in our code. It happened inside `zackees/setup-soldr@v0`, immediately after a successful cache restore:

```
Cache restored successfully
node:events:487
      throw er; // Unhandled 'error' event
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)
Node.js v24.18.0
```

`EPIPE` on a socket write during action setup — infrastructure, not content. The **same commit passed `macos-latest` in the workflow-dispatch dry run minutes earlier**, which is the clearest evidence it is transient.

### Why this matters more than a normal flake

The publish job is gated on all three build jobs succeeding. That gating is correct — a partial release would be worse. But it means **any transient runner hiccup silently converts a tagged release into a no-op**, and the only signal is a red run that someone has to notice and re-run. A tag that produces no crates.io version is a confusing state to leave the repo in.

### Options

- [ ] Add a retry to the setup/build step (e.g. `nick-fields/retry`, or the action's own retry if it has one)
- [ ] Consider whether all three OS build jobs need to gate the publish. The crate is a single source-only publish; the per-OS jobs are validating that it *builds* everywhere. Gating on Ubuntu with macOS/Windows as required-but-retryable may be a better trade — worth a deliberate decision rather than the current implicit one.
- [ ] Make a skipped publish louder: if the tag exists and the publish job did not run, that should be visibly distinguishable from "released fine".

### Workaround for now

`gh run rerun <id> --failed` re-runs just the failed job and lets the publish proceed.

PR #300, opened 2026-09-02
ci: zero-tracking and release assets from cross-built bundles; verify_local --bundle (#277 phases E+F)

zackees:
Implements **#277 phases E and F**. `ci/`, `.github/` and docs only — no C-core or `rust/` paths (CLAUDE.md rule 2).

## Phase E1 — `zero-tracking.yml` loses its `windows-latest` row

That row checked out, built the tree natively with MSVC and *then* ran the A/B: one whole extra Windows VM per matching PR, which is exactly the cost #277 exists to remove. Both halves move into `windows-bundles.yml`'s single `run-windows` job, and they move differently:

| half | before | after |
|---|---|---|
| correctness (`ctest -R zero-tracking`) | Windows row, only on a matching PR | **unconditional, every push, two lanes** — `test-zero-tracking` and `test-zero-tracking-enabled` are registered ctest tests, so `bundle_tests.py` already puts them in every bundle manifest and `run-windows` already replays the whole manifest for `windows-gnu-x64-release` *and* `windows-msvc-x64-release` |
| interleaved A/B timing | Windows row, only on a matching PR | a gated step on `run-windows`, keyed to the same paths |

So the correctness gate got *stronger* while the Windows job count went **down by one**.

A step cannot carry a `paths:` filter, so the gate asks the PR-files API and **fails open** on an API error. It keys on `src/arena.c`, `test/bench-zero-tracking.c` and `.github/workflows/windows-bundles.yml` — deliberately **not** `zero-tracking.yml`, which does not trigger this workflow at all, so keying on it would promise a run the trigger cannot deliver.

Costs stated in `docs/ci-gates.md`: the Windows arm is **clang-cl**, not `cl`; a PR touching only `zero-tracking.yml` exercises the Linux arm alone; and because the gate keys on `windows-bundles.yml`, future CI-only PRs to that file pay ~1 min for the A/B — that is the price of the trigger being provable, and it can be narrowed later.

## Phase E2 — every release binary is cross-built, and `auto-release.yml` says so

**This reverses #277 row E's wording** ("shipped release assets stay built by the platform's own toolchain"). Two owner decisions taken after that row was written force it:

1. **No macOS build or test may run on a native Mac runner** anywhere in this repository (`ci/lint_no_macos_runners.py` enforces it). There is no Apple toolchain available to build an Apple asset with.
2. **Cross-compilation is a product requirement**, not a CI cost measure: mimalloc-pprof ships inside cross-compiled code, so the Linux-cross-built artifact *is* the product.

The second settles Windows, where a native runner does still exist: building the shipped MSVC-ABI binary with `cl` while `windows-bundles.yml` executes a clang-cl bundle would mean **the tested binary and the shipped binary are different binaries**. So the assets come off the same soldr toolchains, the same `cmake/toolchains/soldr-<triple>.cmake` files and the same commit as the executed bundles — the bundles are the evidence for the assets.

`c-unit.yml`'s native `cl` Release `ctest` stays a hard gate (rule 3) and **ships nothing** — there was never a natively-built release asset here to "retain": before this PR `auto-release.yml` shipped only the architecture-independent amalgamation ZIP.

| asset | target | built by | consumer still needs |
|---|---|---|---|
| `mimalloc-pprof-macos-arm64-<tag>.tar.gz` | `aarch64-apple-darwin` | soldr clang 21 + ld64.lld + Apple SDK, on `ubuntu-latest` | — |
| `mimalloc-pprof-macos-x86_64-<tag>.tar.gz` | `x86_64-apple-darwin` | same | — |
| `mimalloc-pprof-windows-x64-gnu-<tag>.zip` | `x86_64-pc-windows-gnu` | soldr mingw-w64 gcc 15 (UCRT) | nothing — `libgcc_s_seh-1.dll` is **shipped in the archive** |
| `mimalloc-pprof-windows-x64-msvc-<tag>.zip` | `x86_64-pc-windows-msvc` | soldr clang-cl + lld-link, `/MD` pinned | the Visual C++ redistributable (`VCRUNTIME140.dll`, `MSVCP140.dll`) |
| `mimalloc-pprof-c-<tag>.zip` | any | source, unchanged | a C compiler |

`cmake --install` (not a hand-picked `cp` list) carries this fork's own `mimalloc/profile.h` and `mimalloc/memory-events.h`, the CMake package, pkg-config, and `mimalloc-redirect.dll` on both Windows lanes — verified on all four lanes, so **no CMakeLists change was needed** (rule 7 escalation not required). The win-gnu archive also ships `libgcc_s_seh-1.dll`, found by `bundle_tests.py`'s own `resolve_runtime_dlls()` import closure: `mimalloc.dll` imports it for `__popcountdi2` and a plain Windows box does not have it, and a missing DLL is a dialog-free `0xC0000135`, not a legible error. Shipping it rather than relinking `-static-libgcc` keeps the shipped binary byte-identical to the one `windows-bundles.yml` tests. Every archive carries a `PROVENANCE.txt` naming commit, target, compiler and the runtime still required.

`build-and-package` drops its `windows-latest` row (review addendum item 5: `xtask check`, `cargo test` and `publish --dry-run` are all platform-independent). `release` `needs:` both build jobs, `release-outcome` reports both, and the rename step hard-fails on fewer than five assets — issue #55 is the "green pipeline published nothing" failure once already.

### How E2 was verified without tagging a release

`auto-release.yml` runs on `workflow_dispatch` and `v*` tag pushes only, so **CI on this PR will not exercise it**. Two substitutes:

- **All four `build-binaries` lanes' `run:` blocks were executed locally on Linux** against real soldr toolchains (`aarch64-apple-darwin`, `x86_64-apple-darwin`, `x86_64-pc-windows-gnu`, `x86_64-pc-windows-msvc`): configure + link-libraries assertion + build + install + DLL closure + arch evidence + `PROVENANCE.txt` + archive. The only step that could not run is the final `zip` on the two Windows lanes — `zip` is absent on this box and present on `ubuntu-latest` (the pre-existing package step already uses it). The gnu lane printed `shipping toolchain runtime: …/libgcc_s_seh-1.dll` and `import closure resolves for 2 PE file(s)`. **No release was tagged and no workflow was dispatched.**
- `ci/tests/test_auto_release_workflow.py` (8 tests): no non-Linux `runs-on`, no `os` matrix dimension, `release` waits for every build job, every uploaded artifact is downloaded, and the build matrix / rename step / `action-gh-release` `files:` list all name the same five assets.

## Phase F — `uv run ci/verify_local.py --bundle <name>`

Fourteen names, exactly the two bundle workflows' build matrices (`macos-{arm64,x64}-{release,debug-full,leak}`, `windows-{gnu,msvc}-x64-{release,debug-full,shared,leak}`). Each runs the same `soldr prepare --target <triple>` (from `rust/`, where the toolchain pin lives), the same toolchain file, the same matrix `cmake` flags and the same `bundle_tests.py` arguments — including `--objdump` / `--dll-search-dir` / `--check-dll-closure` / `--allow-msvc-runtime` — then asserts the resolved `-- Link libraries` line the build job asserts and prints the bundle path, manifest summary and the `run_test_bundle.py` command for the target OS. `--list` prints the table.

It stops **before** execution, the one step Linux cannot do. Everything build-side reproduces in about a minute instead of a push-and-wait cycle: a configure that picked up a host library, a link that lost `__interpose` or its TLS directory, a bundle missing a runtime DLL, a manifest with a leaked absolute path.

The prepared environment is parsed out of `--github-env` and handed to the subprocesses rather than applied to `os.environ`, and a GitHub heredoc-form line is refused rather than mis-parsed.

Verified locally, for real:

```
macos-arm64-release      PASS   38 tests, arm64 Mach-O, Link libraries = pthread
windows-gnu-x64-release  PASS   33 tests, libgcc_s_seh-1.dll shipped
windows-msvc-x64-shared  PASS   30 tests
```

`ci/tests/test_verify_local.py` gains a `BundleTableDriftTests` class that parses both matrices and fails if a bundle name, triple, cmake flag or `bundle_tests.py` argument drifts — the same contract the `--only` config table has had since phase A. The Windows jobs carry no matrix `triple:`, so the test reads it out of the toolchain path their steps pass to cmake rather than hardcoding it.

## Gates

```
pytest ci/tests            229 passed
ruff check / format        clean
pyright 1.1.411 (strict)   0 errors, 0 warnings
lint_no_macos_runners      24 files, no macOS runner labels
```

## Expected red

`macos-bundles.yml`'s `run-macos-x64 (dockurr/macos on Linux)` is **red on `main` too** and unrelated to this PR — the one-time human macOS golden-disk install (`ci/macos_golden_local.sh` → `macos-golden-upload.yml`) is still owed by the owner, and the job deliberately hard-fails on the cache miss rather than skipping itself green. Evidence: run 33654891521 on `main`, same single failing job.

Refs #277.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -1,6 +1,47 @@
 name: auto-release
 
-# Cuts a release of the mimalloc-pprof C amalgamation + crates.io publish.
+# Cuts a release of the mimalloc-pprof C amalgamation + crates.io publish, plus the
+# cross-built binary assets (#277 phase E).
+#
+# WHO BUILDS THE SHIPPED BINARIES -- the decision, and why it is not the one #277's
+# phase table first wrote down:
+#
+#   Every binary release asset is CROSS-BUILT ON LINUX through soldr. Not one of them is
+#   produced by the platform's own toolchain on that platform's own runner.
+#
+# #277 row E originally said "shipped release assets stay built by the platform's own
+# toolchain". Two owner decisions since then overrode it:
+#
+#   1. No macOS build or test may run on a native Mac runner anywhere in this repository
+#      (enforced by ci/lint_no_macos_runners.py, which this workflow is checked by). There
+#      is therefore no Apple toolchain available to build an Apple asset with.
+#   2. Cross-compilation is a PRODUCT requirement, not a CI cost measure: mimalloc-pprof
+#      ships inside cross-compiled code, so the Linux-cross-built artifact IS the product.
+#      Shipping a natively-built binary while testing a cross-built one would mean the
+#      thing under test and the thing shipped are different binaries.
+#
+# So: the same `soldr prepare` toolchains, the same `cmake/toolchains/soldr-<triple>.cmake`
+# toolchain files and the same commit that produce the test bundles executed by
+# `macos-bundles.yml` and `windows-bundles.yml` also produce these assets. The bundles are
+# the evidence for the assets.
+#
+# What that costs, stated rather than hidden:
+#   - macOS assets are built by soldr's clang 21 against soldr's Apple SDK, NOT by Xcode's
+#     clang against Xcode's SDK. Nothing in this repository compiles with Apple's toolchain
+#     any more (docs/ci-gates.md, "macOS").
+#   - The `windows-x64-msvc` asset is MSVC-ABI, but built by clang-cl + lld-link with the
+#     CRT pinned to /MD -- it is not a `cl` binary. c-unit.yml keeps one native `cl`
+#     Release ctest job as the correctness gate for `cl` (CLAUDE.md rule 3); that job ships
+#     nothing, and never did. There has never been a natively-built release asset here to
+#     "retain".
+#   - The `windows-x64-gnu` asset is mingw-w64 gcc 15 UCRT. `mimalloc.dll` imports
+#     `libgcc_s_seh-1.dll` (for `__popcountdi2`), which a plain Windows machine does not
+#     have, so the archive SHIPS that DLL beside it and the job asserts the whole import
+#     closure resolves. Shipping it (rather than relinking `-static-libgcc`) keeps the
+#     shipped binary byte-identical to the one `windows-bundles.yml` tests.
+#   - The `windows-x64-msvc` asset imports VCRUNTIME140.dll/MSVCP140.dll: the Visual C++
+#     redistributable, which is not part of Windows. Recorded in the archive's
+#     PROVENANCE.txt rather than silently assumed.
 #
 # Trigger design (see rust/mimalloc-pprof/vendor/README.md for what gets shipped):
 #   - workflow_dispatch: manual, explicit human/agent decision. Default `dry_run: true`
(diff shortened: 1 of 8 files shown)


zackees:
Opus review: APPROVE. Verified: the zero-tracking A/B gate fired on this PR's own run (21 s, 8+8 samples, artifact present) and its fail-open polarity is the conservative one (API outage runs the A/B); the correctness tests replay in all six bundles; two release lanes reproduced locally against real soldr toolchains with the fork headers, static/shared/object, CMake/pkg-config and `mimalloc-redirect.dll`/`libgcc_s_seh-1.dll` closure; `--bundle` names match the matrices with zero absolute paths. 42/43 green (designed dockur red). Low follow-ups noted on #277: PROVENANCE tag on workflow_dispatch, top-level archive dir, Darwin rpath note, trim MSVCP140 if unused, two-directional arg drift test, `workflow_dispatch` for windows-bundles, and a `dry_run` release run before the next tag.

PR #306, opened 2026-09-02
ci: run the PR benchmarks on demand or daily, never on a docs edit

zackees:
`benchmark-sentinel` ran two 10-minute jobs (ubuntu + windows) on every `pull_request` push touching `rust/`, `src/` or `include/`, and `windows-bundles` ran a third sentinel on the single shared Windows VM on the same event. Nothing consumes those numbers — they are informational by construction (#170/#171: a hosted runner is not a stable reference host) and no gate reads them — so the per-push cadence spent ~20 runner-minutes per push, in the same hosted pool the correctness lanes queue for. A markdown-only edit under `rust/` paid it too.

## After

| when | measures? |
|---|---|
| docs-only PR (`**/*.md`, `**/*.txt`, `docs/**`) | not even queued |
| ordinary PR push | no — job reports `skipped` |
| PR labelled `benchmark` | yes (the `labeled` event starts it, no dummy commit) |
| `workflow_dispatch` | yes |
| daily cron (`43 6 * * *`) | yes — keeps the trend line |

`windows-bundles`' `PR benchmark sentinel (MSVC, cross-built)` step and its artifact upload take the same label condition. Its Linux-side cross-build stays unconditional: it feeds an unconditional `download-artifact`, and time on the shared Windows VM — not a cargo build on the Linux builder — is the scarce thing.

The publication workflows (`benchmark-stats` daily; `benchmark-memory`, `benchmark-latency`, `benchmark-scaling` weekly) are untouched: they are already dispatch-plus-cron only and never ran on a pull request.

### DIFF
--- a/.github/workflows/benchmark-sentinel.yml
+++ b/.github/workflows/benchmark-sentinel.yml
@@ -4,14 +4,38 @@ name: benchmark-sentinel
 # Produces JSON + HTML report uploaded as a CI artifact.
 # Hosted-runner results are labelled informational — not accepted for
 # small-timing claims (requires stable reference host per #170).
+#
+# ON DEMAND, OR ONCE A DAY -- NOT ON EVERY PUSH. This used to run two 10-minute
+# jobs on every pull_request push that touched rust/, src/ or include/. Nobody
+# reads the result of most of those: the numbers are informational by
+# construction (#170/#171 -- a hosted runner is not a stable reference host) and
+# no gate consumes them, so a per-push cadence bought a trend nobody plots at the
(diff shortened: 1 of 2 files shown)


zackees:
Reviewed: workflow-only. PR benchmark measurements now run only on a `benchmark` label (`labeled` event, no dummy commit), `workflow_dispatch`, or the daily cron; docs edits excluded by path filter; the Windows sentinel step follows the same label gate while its Linux cross-build stays for the unconditional download. Consistent with the owner's request to stop the CI queue from clogging. Merging.

Issue #307, opened 2026-09-03
ci(c-unit): build every configuration once, then run all tests at once — stop rebuilding per job


zackees:
## Problem (owner, 2026-09-03)

`c-unit.yml` is laid out as 16 jobs that each check out, configure, and **rebuild the library** for one configuration, then run one slice of the tests (`ctest`, `ctest-debug-full`, `ctest-shared`, `ctest-guarded`, `ctest-pprof-off`, `ctest-musl`, `memory-gate`, `diagnostic-gates`, `isa-baseline`, `bundle-roundtrip`, `bun-surface`, the three MinGW rows, …). Run 33699111919: 20 jobs, most of their wall time is the same build repeated. Owner direction: **"All the tests need to be created all at once, all the time, and then the runners need to run all the tests all at the same time, not a bunch of tests that rebuild the product over and over."**

## Target layout

Two stages, same model as `windows-bundles.yml`/`macos-bundles.yml` (#277):

1. **Build stage — every configuration, once, in parallel.** A `build` matrix on Linux (one job per configuration: `release` (`MI_PPROF=ON`), `pprof-off`, `debug-full`, `guarded` (Debug + `MI_GUARDED`), `shared`, `musl` (alpine container), `memory-gate-leak` (`MI_BENCH_INJECT_LEAK`), `isa-portable`/`isa-arch` (build-only artifacts), plus the Windows native `cl` Release config on `windows-latest` (rule 3: the retained native MSVC gate stays built by `cl` — but built ONCE). Each job builds all test binaries for its configuration and uploads a test bundle (`uv run ci/bundle_tests.py`) — same bundle format Phase A defined. No job in this stage runs tests except what is inseparable from building (none).
[...]

PR #309, opened 2026-09-03
ci: make the memory gate measure the allocator, not the runner (#298)

### LINKED ISSUE #298: ci: ubuntu memory-gate is noise-bound — re-baseline from a main run on the current runner and set tolerance from measured spread

zackees:
Closes #298.

The ubuntu `memory-gate` job flipped red on commits that changed no allocator code, and
its own printout warned that the observed spread exceeded the tolerance it was compared
under. This makes the measurement deterministic, then sets the tolerance from what was
measured.

### What the artifacts said

`memory-gate-ubuntu-latest` for the last twelve `main` runs of `c-unit.yml`, against a
22.3 MB baseline with 25.6 MB allowed:

| run | min-of-8 | median | within-run spread |
|---|---|---|---|
| 33618717188 | 26.7 | 29.6 | 32.6% |
| 33629562858 | 24.7 | 27.4 | 19.0% |
| 33631613324 | 28.0 | 29.2 | 15.7% |
| 33635722154 | 26.0 | 28.0 | 19.6% |
| 33636560081 | 22.3 | 26.9 | 30.0% |
| 33646494496 | 24.3 | 28.2 | 44.4% |
| 33647332787 | 24.5 | 26.5 | 17.6% |
| 33654892179 | 27.9 | 31.2 | 24.4% |
| 33660325957 | 26.5 | 27.6 | 18.5% |
| 33689409217 | 22.7 | 25.6 | 25.6% |
| 33692432919 | 25.5 | 27.9 | 28.2% |
| 33699070304 | 27.8 | 30.2 | 19.8% |

min-of-8 spans 25.6% across runs and the median 21.9%. **No statistic was going to rescue
this** — the answer to "are the tolerances wrong?" is that the tolerance was not the
problem. `profiler_arena_mb`, `purged_gb` and the liveness counters are constant in every
one of those 96 records, so the profiler arena and the P7a/P7b purge machinery are ruled
out; the variance is entirely in peak RSS.

The same twelve runs' `memory-gate-windows-latest` artifacts: min-of-8 51.2–57.0 MB
(11.3% across runs), within-run spreads 3.9–17.5%. Noisy, but it passed — confirming that
only ubuntu was actually flaky, and that the Windows baseline had itself been recorded
from a measurement noisier than any tolerance worth having (9.9%).

### Where the noise was

Per-scenario decomposition on an idle host pinned to 4 CPUs: **scenario 1's 8-thread churn
alone ranged 14.1 → 24.6 MB run to run**, and every later scenario added ≤ 2 MB on top.
Peak RSS is a high-water mark, so the gated number was *how many churn workers the
scheduler happened to overlap*.

### The fix is in the workload, not the threshold

`test/test-memory-gate.c`: the churn workers rendezvous on a portable condition-variable
barrier (allocate → all wait → churn → all wait → free), so "all N live at once" is the
only reachable state and the peak is the structural cost of N live sets. Scenario 3's
`while (ready == 0) {}` handoff becomes the same barrier — six spinning consumers on a
4-vCPU runner preempt the producers they wait on. Wall time roughly halved (0.26 / 0.47 /
0.52 s → 0.27 / 0.18 / 0.18 s, three runs each, `taskset -c 0-3`).

Three independent `ubuntu-latest` runner VMs (`nproc=4`, THP=`always`), 16 runs each,
identical object code:

| lane | min | median | max | spread |
|---|---|---|---|---|
| new workload, VM A | 58.0 | 58.3 | 58.5 | **0.9%** |
| new workload, VM B | 58.0 | 58.5 | 58.5 | **0.9%** |
| new workload, VM C | 58.0 | 58.3 | 58.6 | **1.0%** |
| old workload, same three VMs | 26.4 / 27.4 / 25.5 | | | 26.1% / 24.5% / 23.1% |

min-of-16 is 58.0 MB on all three. `taskset -c 0-1` (0.5%), `taskset -c 0` (0.7%) and
`MIMALLOC_SCAVENGER=0 MIMALLOC_PURGE_HOLES=0` (1.2%) change nothing — **the spread
collapsed because of the barrier, not because of any knob.** An unpinned 16-core
workstation reads 58.2–58.3 MB, the same number as the 4-vCPU runner, where the old
workload read 58 MB unpinned vs 23 MB pinned.

### Statistic, tolerance, baselines

- **Statistic: still min-of-8.** One-sided noise makes the minimum right, but it is no
  longer load-bearing — min-of-8 and min-of-16 of this distribution are the same number.
  `RUNS_EXPECTED` stays 8 for that reason (measured, not assumed).
- **`PEAK_TOLERANCE` 0.15 → 0.05.** 5× the largest observed spread, 2.9 MB of headroom on
  58.0 MB. The measurement supports 0.03; the rest is reserved for runner-image drift,
  which three runs on one image cannot bound. The comment says what to measure before
  tightening.
- **Control margin: +42.1%, 8.4× the tolerance.** CI confirms it on both gated metrics:
  run 33701827771 reports `peak_rss regressed: 82.5 MB vs baseline 58.0 MB (+42.2%)` and
  `peak_commit regressed: 113.7 MB vs baseline 74.4 MB (+52.8%)`. The rule is that the
  control must fire by ≥2× the tolerance, and a test asserts that ratio, so raising the
  tolerance past half the measured control margin is a failing test rather than a
  judgement call. Note the control no longer fires at ~3× the *baseline* as it used to:
  the injected leak is the same ~32 MB, but the structural peak it must clear went from
  ~22 MB to 58 MB. `MI_BENCH_INJECT_LEAK` is set in the workflow, so raising it belongs
  with #307's `c-unit.yml` rewrite rather than here.
  Leak injection now `memset`s the whole block: one measured batch had a control run land
  within 1% of the clean baseline, i.e. the control came within one sample of silently not
  firing.
- **`linux-pprof1.json` → 58.0 MB** (min-of-16, 0.9% spread) from run 33700518749.
- **`windows-pprof1.json` → 74.4 MB** (min-of-8, **0.1% spread**, from 17.5%) from run
  33701098707, reproduced exactly by run 33701827771 on a second Windows runner. That one
  source change collapses the spread on two unrelated kernels and two different gated
  metrics is the strongest evidence the diagnosis was right.
- **`macos-pprof1.json` deliberately untouched and now stale**: it records the native
  Xcode arm64 lane #277 phase B2 deleted, and nothing compares against it (the dockur
  x86_64 lane keys on `macos-x86_64-soldr-clang-21-pprof1.json` and bootstraps its own).
  Documented as such in `docs/ci-gates.md`.
- A new test asserts **every** committed baseline's own recorded `baseline_spread_pct` is
  below the tolerance it will be compared under. It is what caught the stale Windows one.

### Verification

`memory-gate (ubuntu-latest)` and `memory-gate (windows-latest)` both **pass**, positive
controls included, on runs 33701098707 and 33701827771. Five independent `ubuntu-latest`
VMs read 58.0–58.2 MB (0.7–1.0% spread); two Windows runners read 74.4 MB (0.1%).
`run-macos-x64 (dockurr/macos on Linux)` is red for the pre-existing reason documented in
`docs/ci-gates.md` (no golden disk yet), unrelated to this change.

Follow-up, not done here: `gh run rerun` on a `main` run only re-runs main's workload
against main's baseline, so it proves nothing pre-merge — worth doing once this lands.

### Coordination

**No workflow file is touched.** Everything is in `test/test-memory-gate.c`,
`ci/memory_gate.py`, `ci/memory-baselines/`, `ci/tests/` and `docs/ci-gates.md`, so #307's
rewrite of `c-unit.yml` carries this without a conflict. The C-core workload change is its
own commit (rule 2).

### DIFF
--- a/ci/memory-baselines/linux-pprof1.json
+++ b/ci/memory-baselines/linux-pprof1.json
@@ -1,23 +1,23 @@
 {
-  "baseline_runs": 4,
-  "baseline_spread_pct": 5.8,
+  "baseline_runs": 16,
+  "baseline_spread_pct": 0.9,
   "counters": {
-    "pages_end": 1,
-    "pages_start": 1,
+    "pages_end": 2,
+    "pages_start": 2,
     "theaps_end": 1,
-    "theaps_start": 0,
+    "theaps_start": 1,
     "threads_end": 1,
     "threads_start": 1
   },
   "gated_metric": "peak_rss",
   "inject_leak": 0,
   "mi_pprof": 1,
-  "peak_commit_mb": 49.7,
-  "peak_mb": 22.3,
-  "peak_minus_profiler_mb": 22.3,
-  "peak_rss_mb": 22.3,
+  "peak_commit_mb": 61.4,
+  "peak_mb": 58.0,
+  "peak_minus_profiler_mb": 58.0,
+  "peak_rss_mb": 58.0,
   "platform": "linux",
   "profiler_arena_mb": 0.0,
-  "purged_gb": 0.06,
+  "purged_gb": 0.1,
   "schema": 1
 }
--- a/ci/memory-baselines/windows-pprof1.json
+++ b/ci/memory-baselines/windows-pprof1.json
@@ -1,23 +1,23 @@
 {
-  "baseline_runs": 4,
-  "baseline_spread_pct": 9.9,
+  "baseline_runs": 8,
+  "baseline_spread_pct": 0.1,
   "counters": {
-    "pages_end": 1,
+    "pages_end": 0,
     "pages_start": 0,
     "theaps_end": 1,
-    "theaps_start": 0,
+    "theaps_start": 1,
     "threads_end": 1,
     "threads_start": 1
   },
   "gated_metric": "peak_commit",
   "inject_leak": 0,
   "mi_pprof": 1,
-  "peak_commit_mb": 52.5,
-  "peak_mb": 52.5,
-  "peak_minus_profiler_mb": 52.5,
-  "peak_rss_mb": 16.2,
+  "peak_commit_mb": 74.4,
+  "peak_mb": 74.4,
+  "peak_minus_profiler_mb": 74.4,
+  "peak_rss_mb": 22.6,
   "platform": "windows",
   "profiler_arena_mb": 0.0,
-  "purged_gb": 0.06,
+  "purged_gb": 0.1,
   "schema": 1
 }
--- a/ci/memory_gate.py
+++ b/ci/memory_gate.py
@@ -10,12 +10,19 @@ Windows commit report different numbers for identical allocator behavior, and ma
 compresses memory so its resident figure falls under pressure while a leak grows. Only
 same-OS deltas are meaningful.
 
-Peak memory is *not* a low-variance signal, so both commands take several runs of the
-same binary and use the **minimum** observed peak. Noise on a shared CI runner only ever
-adds memory (another tenant's page cache, a straggler thread still winding down), so the
-minimum is the estimator closest to the allocator's true high-water mark, and it is far
-more stable than any single run. The spread across runs is printed every time -- if it
+Both commands take several runs of the same binary and use the **minimum** observed peak.
+Noise on a shared CI runner only ever adds memory (another tenant's page cache, a
+straggler thread still winding down), so the minimum is the estimator closest to the
+allocator's true high-water mark. The spread across runs is printed every time -- if it
 ever approaches the tolerance, the tolerance is wrong and the printout says so.
+
+The minimum is not what makes this gate stable, though; #298 established that no
+statistic could rescue the old workload (min-of-8 across twelve consecutive `main` runs
+spanned 22.3-28.0 MB, the median 25.6-31.2 MB), because the peak was set by how many of
+test-memory-gate's churn workers the runner's scheduler happened to overlap. The fix was
+in the workload, not here: test/test-memory-gate.c now rendezvouses those workers, so the
+peak is the structural cost of N live sets rather than a scheduler coin flip. The minimum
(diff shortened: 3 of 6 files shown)


zackees:
Opus review: APPROVE. Recomputed from artifacts: 12 main runs reproduce the PR's distribution table; the fix is in the workload (CV barrier so all 8 churn workers hold their live set simultaneously; scenario-3 spin → 2-party barrier), collapsing spread ubuntu 26% → 0.7–1.0% and windows 17.5% → 0.1%; `PEAK_TOLERANCE` 0.05 is 5× the largest observed spread with better absolute sensitivity than before; leak control fails at +42% (8.4× tolerance) on Linux and +52.8% on Windows; baselines byte-identical to `update` on the cited artifacts; rule 2 honoured (test change is its own commit). Non-blocking items forwarded to #307 (workflow comment, stale macOS baseline). Merging.

PR #310, opened 2026-09-03
fix: gcc-ASan test-api crash (#301) and toolchain-agnostic local asan config

### LINKED ISSUE #301: ASan: test-api SEGV in src/arena.c (~1204) on main under gcc -fsanitize=address RelWithDebInfo
Found during the PR #299 review, reproduced on `origin/main` (77143426) independent of that PR: with a gcc ASan build (`-fsanitize=address`, RelWithDebInfo, `MI_TRACK_ASAN`), `test-api` segfaults at `src/arena.c:1204` (arena.c:1216 on the P7a branch). The CI `asan.yml` job uses clang and is green, so this is either a gcc-ASan-specific interaction (e.g. ASan's malloc interposition vs mimalloc override) or a real bug that clang's instrumentation hides. Reproduce: `cmake -S . -B out/asan-gcc -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMI_TRACK_ASAN=ON -DCMAKE_C_COMPILER=gcc -DCMAKE_C_FLAGS=-fsanitize=address && cmake --build out/asan-gcc && out/asan-gcc/mimalloc-test-api`. Note: local `verify_local.py`'s `asan` config fails to configure on this workstation (clang lacks `sanitizer/asan_interface.h`), which is why agents fall back to gcc ASan — worth making the asan config toolchain-agnostic too. Refs #264 (P7a review), #277.

zackees:
Closes #301.

## Root cause: not gcc, not ASan

`RelWithDebInfo -DMI_PADDING=1` — no sanitizer, **and it crashes identically under gcc and clang** — reproduces the SEGV. `MI_TRACK_ASAN` merely implies `MI_PADDING` (`types.h:105`).

ASan/gdb evidence: `mi_free_size(p, 1024)` → `mi_free_small(p)` → SEGV at `arena.c:1223`, `heap->subproc` with `heap == NULL`. `p == 0x7bffb3b60000` is itself 64 KiB-aligned, and the "page" the crash dereferences is the caller's own zeroed block:

```
$1 = {xthread_id = 0, free = 0x0, used = 0, ..., heap = 0x0, ...}   # p, not an mi_page_t
```

`MI_OPT_FREE_SMALL` makes `mi_free_small` find the page as `_mi_align_down_ptr(p, MI_SMALL_PAGE_SIZE)`, valid only while the page meta sits in front of the slice. `mi_arenas_page_alloc_fresh` decided that on `block_size > MI_SMALL_SIZE_MAX` — but the small path is *entered* on the **requested** size, and padding makes the two diverge: `mi_good_size(1024) == 1280`. A 1009..1024 byte request took the small path while its meta went to `pages_meta`. Fix: use the bound `mi_free_small`'s own assert states, `mi_good_size(MI_SMALL_SIZE_MAX)` — a literal no-op without padding.

**Why CI was green:** `asan.yml` built `Debug`, and CMakeLists auto-enables `MI_OPT_FREE_SMALL` **only when `MI_DEBUG` is off**. The two ingredients could never coexist in that lane.

## Second defect in the same configuration

With the SEGV fixed, upstream's own `mi_urealloc_invalid` fails in any non-debug `MI_PADDING` build. A failed padding decode makes `mi_page_usable_size_of` return 0 ("size can be zero if the padding is corrupted"); `mi_theap_realloc_zero_ex` then reallocated and **freed `p`** — an interior pointer onto the page free list. Now returns `NULL` without freeing, as its own header comment promises and as `MI_DEBUG` already does. No valid block reports usable size 0 — verified by exhaustively probing `mi_malloc`/`zalloc`/`calloc`/`realloc` 0..4096 and `mi_*_aligned`/`mi_malloc_aligned_at` 0..2048 × alignments 1..8192 across padding/non-padding, debug/release: zero hits.

## Provenance (measured against `dev3` HEAD, not inferred)

Both reproduce on **stock upstream `6def7be9`** — this fork's pin, and Bun's mimalloc merge-base — with zero fork changes.

A stock `34fbd7e7` run proves nothing about either: `include/mimalloc/types.h:68` there carries an unconditional `#define MI_PADDING 0` (added by `cda82f99`) *above* the `MI_TRACK_ASAN`/`MI_SECURE>=3` auto-enable, so `-DMI_PADDING=1` is clobbered with only a redefinition warning and **`dev3` HEAD ships no block padding in any configuration** — itself an upstream defect worth reporting.

Neutralize that line (padding then really engages: `MI_PADDING_SIZE == 8`, `mi_good_size(1024) == 1280` — the divergence is *unchanged* at dev3) and build with `-DMI_OPT_FREE_SMALL=ON`; `test-api` is 49 ok / 1 failed:

| Defect | `6def7be9` | `34fbd7e7` | Upstreamable |
|---|---|---|---|
| **4** `mi_free_small` reads the caller's block as a page | reproduces | does **not** reproduce (small-page free path reworked by `82fd55ff`/`fe5a54c4`/`eab1ebe6`; not bisected) | no |
| **5** `mi_realloc` frees an interior pointer | reproduces | **still fails** `mi_urealloc_invalid` | **yes** |
| `MI_PADDING` unreachable (`types.h:68`) | n/a | present | **yes** |

Recorded in `docs/upstream-bugs.md` (bugs 4–5) and `docs/upstreaming.md`.

## Cost in the default build: zero

The `arena.c` bound is `#if MI_PADDING`-selected and the `alloc.c` check is `#if (MI_PADDING || MI_GUARDED)`-guarded, so in a Release non-padding build **`arena.c.o` and `alloc.c.o` disassemble byte-identically to `main`**, `mi_urealloc` included (224 instruction lines, zero diff).

## Regression coverage

| Lane | Before | After |
|---|---|---|
| `asan.yml` `clang, Debug` (+`MI_DEBUG_FULL`) | hard gate | unchanged, hard gate |
| `asan.yml` `gcc, RelWithDebInfo` | — | **new**, hard gate — the reported config |
| `asan.yml` `clang, RelWithDebInfo` | — | **new**, hard gate — asymmetry would itself be a finding |
| `verify_local.py` `asan` | clang-only; could not configure where clang lacks `sanitizer/asan_interface.h` | probes clang → gcc → SKIP-with-reason; runs both build types |
| `ci/tests/test_verify_local.py` | a matrixed `-D` flag read as the literal token `-DCMAKE_BUILD_TYPE=${{` | expands `${{ matrix.* }}` first, so matrixed flags stay drift-checked |

## Verification (this workstation)

- `uv run ci/verify_local.py --keep-going` → **11/11 PASS** (the `asan` config picked gcc — clang here has no compiler-rt headers — and ran Debug + RelWithDebInfo).
- gcc ASan `RelWithDebInfo`: `test-api` 46 ok / 0 failed; full `ctest` **45/45**; UAF positive control still fires (`use-after-poison`).
- Same with `-DMI_DEBUG=2` so `mi_assert_internal` is armed on the widened meta-in-front set: **45/45**.
- Non-padding `RelWithDebInfo` unchanged: `test-api` 47 ok / 0 failed.
- Zero-usable-size probe extended through every page class (8K/64K/512K/1M/8M/64M ±1, aligned and `aligned_at`, plus realloc ping-pong across the small/large boundary): still no allocation reports 0.
- `ruff` + `pyright --strict` + `pytest ci/tests` (237 passed); `ci/check_doc_snippets.py` PASS; `xtask check` green after the amalgamation regen.

**Pre-existing red on `main` at the branch point (`af947dff`), not from this PR:** `memory-gate (ubuntu-latest)` (fails on main's own run 33699070304) and `run-macos-x64 (dockurr/macos on Linux)` (documented in `docs/ci-gates.md` — the golden disk does not exist yet).

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -25,11 +25,36 @@ concurrency:
 
 jobs:
   asan:
-    name: address sanitizer
+    # Issue #301. One row is not enough. `MI_TRACK_ASAN` implies `MI_PADDING`, and
+    # CMakeLists auto-enables `MI_OPT_FREE_SMALL` (hence `MI_PAGE_META_ALIGNED_FREE_SMALL`)
+    # only when `MI_DEBUG` is off -- so a Debug-only ASan lane can never reach the
+    # `mi_free_small` fast path at all, and sat green through a NULL-deref SEGV in
+    # `mi_arenas_page_free_ex` that any release-type padding build hits in `test-api`'s
+    # `free_small1`. The RelWithDebInfo rows are that missing coverage. gcc is in the matrix
+    # because gcc is what the bug was reported against, and clang stays because the bug is
+    # compiler-independent -- if a row ever goes red on one compiler only, that asymmetry is
+    # itself the finding.
+    name: address sanitizer (${{ matrix.cc }}, ${{ matrix.build_type }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - cc: clang
+            cxx: clang++
+            build_type: Debug
+            extra: -DMI_DEBUG_FULL=ON
+          - cc: gcc
+            cxx: g++
+            build_type: RelWithDebInfo
+            extra: ''
+          - cc: clang
+            cxx: clang++
+            build_type: RelWithDebInfo
+            extra: ''
     env:
-      CC: clang
-      CXX: clang++
+      CC: ${{ matrix.cc }}
+      CXX: ${{ matrix.cxx }}
     steps:
       - uses: actions/checkout@v4
 
@@ -39,8 +64,8 @@ jobs:
       - name: Configure (and prove ASan actually enabled)
         run: |
           set -euo pipefail
-          cmake -B build-asan -DCMAKE_BUILD_TYPE=Debug \
-                -DMI_PPROF=ON -DMI_DEBUG_FULL=ON -DMI_TRACK_ASAN=ON 2>&1 | tee configure.log
+          cmake -B build-asan -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+                -DMI_PPROF=ON ${{ matrix.extra }} -DMI_TRACK_ASAN=ON 2>&1 | tee configure.log
           if ! grep -q "Compile with address sanitizer support (MI_TRACK_ASAN=ON)" configure.log; then
             echo "::error::MI_TRACK_ASAN did not survive configure -- the sanitizer is NOT enabled."
             grep -i "sanitizer" configure.log || true
--- a/ci/tests/test_verify_local.py
+++ b/ci/tests/test_verify_local.py
@@ -30,6 +30,7 @@ WORKFLOWS = ROOT / ".github" / "workflows"
 
 DFLAG_RE = re.compile(r'-D([A-Z_][A-Z0-9_]*)=("\$pprof"|\$pprof|[^\s"\'\)]+)')
 SCRIPT_RE = re.compile(r"ci/[A-Za-z0-9_./-]+\.py")
+MATRIX_REF_RE = re.compile(r"\$\{\{\s*matrix\.([A-Za-z0-9_-]+)\s*\}\}")
 
 
 def load_workflow(name: str) -> dict[str, Any]:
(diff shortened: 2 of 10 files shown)


zackees:
Opus review: APPROVE after fixes. Both defects reproduced on stock upstream 6def7be9 (padding release build: `mi_free_small` misroutes 1009–1024 B frees → SEGV; `mi_realloc` frees an interior pointer on undecodable padding). Fixes are guarded (`#if MI_PADDING` / `#if (MI_PADDING || MI_GUARDED)`) and Release non-padding `alloc.c.o`/`arena.c.o` disassemble byte-identically to main. Docs corrected: dev3 disables padding entirely (`MI_PADDING 0` in types.h), so bug 4 is only hidden there by the free-path rework and bug 5 remains upstreamable. `asan.yml` now gates gcc and clang RelWithDebInfo alongside clang Debug; `verify_local` asan probes the toolchain. Merging; closes #301.

PR #311, opened 2026-09-03
ci(c-unit): build every configuration once, run all tests at once (#307)

### LINKED ISSUE #307: ci(c-unit): build every configuration once, then run all tests at once — stop rebuilding per job

zackees:
Closes #307.

`c-unit.yml` was 16 jobs that each checked out, configured and **rebuilt the library** for one configuration before running one slice of the suite. It is now two stages: every configuration is compiled exactly once, in parallel, and one job per OS executes everything.

## Layout

**Stage 1 — build (one job per configuration, none of them runs a test)**

| job | rows | produces |
|---|---|---|
| `build` | `release`, `pprof-off`, `debug-full`, `guarded`, `shared`, `musl`, `musl-pprof-off` | tarred test bundle + `ctest --show-only=json-v1` |
| `build` | `memory-gate-leak` | a bundle that is never executed — only its `mimalloc-test-memory-gate`, as the gate's positive control |
| `build` | `isa-portable`, `isa-arch`, `diag-pprof-on`, `diag-pprof-off` | `libmimalloc*.a` (+ `compile_commands.json`) for the run stage to scan |
| `build-bun-objects` | glibc, musl | Bun probe objects, so the run stage never recompiles `src/static.c` |
| `build-windows-native` | — | the retained `cl` Release bundle |

**Stage 2 — run**

- `run-linux`: unpacks every Linux bundle and runs the five glibc ones in **one wave** (`ci/run_test_bundle.py --bundles … --jobs 4`), with guarded's `MIMALLOC_GUARDED_SAMPLE_RATE=1` pass as an `--env-variant` **scoped to that bundle** instead of a whole second `ctest` run. Then, as separate steps so each has the machine to itself: the musl bundles inside `alpine:3.20`, the memory gate + leak control, the diagnostic/ISA scans, the Bun surface link, and the coverage comparison.
- `run-windows-native`, still named **`ctest (windows-latest)`**: runs the `cl`-built bundle. Rule 3 gate preserved by name and by toolchain; only the compile and the run are now separate jobs.
- `isa-baseline (ubuntu-24.04-arm)` stays a self-contained build+scan job — `check_isa_baseline.py` disassembles with the host's single-target `objdump`, so an x86_64 runner has nothing to hand it.

## No test may be dropped

`ci/bundle_coverage.py` now reads a JUnit candidate side, so for every configuration it compares **every name `ctest --show-only` registered in the build job against every name the run stage actually executed**. The reference comes from ctest, the candidate from the run; neither is derived from the other. `bundle-roundtrip`'s pass/fail equivalence check moved to `ci/verify_local.py --only bundle`, where a build tree exists to compare against — CI's run stage has no build tree at all, so the RPATH escape hatch that job's `mv` guarded against cannot exist there.

## Concurrency must not change what a test asserts

The RSS- and deadline-sensitive tests carry `RUN_SERIAL TRUE` in `CMakeLists.txt` (ctest's own word for it); `bundle_tests.py` lowers it to a `serial` flag and the runner holds those back and runs them **alone**, after the wave, across all bundles. A test that passes serially and fails under the wave belongs in that list with a comment saying which property it depends on — not behind a retry. `ctest -j` honours the same declaration locally.

## Job counts

| | before (run 33699111919) | after |
|---|---:|---:|
| jobs | 20 | 17 (12 `build` rows + 2 bun + 1 windows build + `run-linux` + `run-windows-native` + `isa-baseline (arm)`) |
| `windows-latest` jobs | 7 | 2 |

The three MSYS2 win-gnu rows and the `windows-latest` rows of `ctest-debug-full`/`ctest-shared`/`memory-gate` are deleted. Their dated TODOs (2026-09-02) said "delete after ten pushes have compared this against `run-windows` in windows-bundles.yml"; there have been **19 `c-unit` runs on `main` since**, with `windows-bundles.yml`'s `run-windows` green throughout. win-gnu stays gated in `windows-bundles.yml` (CLAUDE.md rule 3); `docs/ci-gates.md` and rule 3 record both facts. **Flagging this for owner review** — it is what #307's target layout asks for, and the window has demonstrably elapsed, but it is the one judgement call in this PR.

## Measurements

**Before — run 33699111919** (20 jobs; note two of them were abnormally short: `memory-gate (ubuntu-latest)` failed at 38s and `bun-surface (glibc)` finished in 8s, so the "before" is slightly flattering):

- raw runner-minutes: **51.0**
- billed runner-minutes (Windows ×2): **74.4**
- wall clock, first job start → last job end: **9.75 min**

**After — run [33707744137](https://github.com/zackees/mimalloc-pprof/actions/runs/33707744137), green** (20 jobs):

| | before (33699111919) | after (33707744137) |
|---|---:|---:|
| jobs | 20 | 20 |
| raw runner-seconds → minutes | 51.0 | **23.9** (−53 %) |
| **billed** (GitHub rounds each job up to the next minute, Windows ×2) | **86** | **41** (−52 %) |
| wall clock (incl. queue) | 9.75 min | 13.5 min |
| critical path of *work* | ~7.6 min (`ctest-guarded`) | **9.25 min** — `build (debug-full)` 34 s → `run-linux-serial` 509 s → `coverage` 12 s |
| `windows-latest` jobs | 7 | 2 |

The raw-seconds row is what the jobs actually consumed; the **billed** row is what the
account is charged, since GitHub rounds every job up to a whole minute and doubles
Windows. The old layout's many short jobs each paid a full minute, so per-job rounding
costs it more than the new one: 86 → 41 minutes.

`run-linux` (the 4-wide wave + the quiet-machine gates) is 5.6 min and `run-linux-serial`
(the `RUN_SERIAL` group, one test at a time) is 8.5 min; they run **concurrently on two
runners**, so the serial group no longer sits on the critical path. Measured on the
single-job version of this PR (run 33704383536) the serial group was 508 s of 2199
single-threaded test-seconds — 23 %, ~8.5 of that job's 16 min — which is exactly the
8.5 min the split moved off the path. The serial list itself is unchanged.

Net: **critical path within ~1.5 min of the old layout, at 37 % of the billed
runner-minutes**, with every configuration compiled exactly once.

**Local `--like-ci` measurement** (16-core box, `--jobs 12`, 5 bundles + the guarded variant = 278 test runs):

- single-threaded test time across all bundles: **2519 s**
- wall clock for the whole wave: **567 s**
- serial group: 57 items, **363 s = 14 % of test-seconds** — and that part does not divide by `--jobs`

Set expectations honestly: the parallel wave divides by `--jobs`, the serial group does not, so on a 4-vCPU runner `run-linux` is the new critical path and is expected to be **longer in wall-clock than the old 9.75 min while costing roughly half the runner-minutes**. That is the trade #307 asks for ("stop rebuilding the product over and over"). Two levers if the owner wants wall-clock back, neither taken here:

1. a second `run-linux-serial` job — a separate VM is exclusive by construction, so the serial group stops being additive;
2. guarded's `sample-rate-1` variant is ~35 % of the test-seconds on its own and could be its own row.

## Verified locally

`uv run ci/verify_local.py --like-ci` → PASS (builds all 10 Linux configurations once, replays every bundle in one wave, then coverage + memory gate + leak control + diagnostic/ISA scans + both halves of the Bun surface check). `pytest ci/tests` 273 passed, `ruff check`/`format`, `pyright` strict clean, `ci/lint_no_macos_runners.py` clean.

`--like-ci` earned its keep on the first run: it caught a latent bug in `run_test_bundle.py`'s JUnit writer. mimalloc's `MIMALLOC_VERBOSE=1` output is ANSI-coloured, ESC (0x1b) is illegal in XML 1.0 with no legal escape, and nothing had ever parsed those files back — so `guarded.xml` failed the new coverage step with "not well-formed (invalid token)" instead of reporting on coverage. Fixed in its own commit, with a test.

The musl claim was measured rather than assumed: a musl-linked executable from this tree requests `/lib/ld-musl-x86_64.so.1`, so a glibc host cannot run it; `python3` + `libstdc++` is the complete runtime set, confirmed by running the whole suite in a container with nothing else installed.

## Notes for review

- `ci/memory_gate.py`, `ci/memory-baselines/` and `test/test-memory-gate.c` are **untouched** — #298 is revisiting the tolerance concurrently and must not have to fight this PR. The run job calls the existing `check`/`control` CLI with the same 8 runs.
- `ci/tests/test_verify_local.py`'s drift guard now parses matrix `include:` rows as well as `run:` blocks. Without that, moving the cmake flags into `c-unit.yml`'s build matrix would have made the guard silently vacuous for the one job that decides what gets built at all.
- `build-windows-native` producing a bundle from a **native Visual Studio multi-config tree** is the one piece not reproducible on this box; it asserts `mimalloc.dll` + `mimalloc-redirect.dll` are in the bundle and that the manifest carries no absolute path, so a layout surprise fails legibly rather than as an 0xC0000135 on the run job.


zackees:
Opus review: APPROVE. Old→new mapping verified from the green run (every registered test executed per config; 278 glibc + 86 musl; guarded variant scoped to its bundle; serial group complete and on its own VM; native `cl` Release ctest retained as the hard gate under its original check name; deleted informational Windows rows all have windows-bundles equivalents; leak control fires at 34× tolerance; baselines untouched). Timings: raw runner-minutes 51.0 → 23.9 (−53%), GitHub-billed ≈86 → ≈41 (−52%), critical path ~9.25 min. Doc fixes landed at `2d8d152d`. Merging; closes #307.

PR #321, opened 2026-09-03
feat(rust): bind every fork C API and document the C/Rust surface table

zackees:
Answers "does the hole-punch API — and every other API — exist in Rust?", closes the gaps it found, and adds the two gates that stop them reopening.

## The answer, short

The hole-punch API **was** in Rust, all of it except `mi_purge_holes_report`. Three other areas were not:

| Area | Before | After |
|---|---|---|
| `include/mimalloc/memory-events.h` | **nothing bound** except the three `mi_unwrapped_*` | fully bound + `memory_events` safe module |
| `mi_option_*` (incl. all 13 fork options) | **nothing bound**, no `mi_option_t` mirror | fully bound + `options` safe module |
| `include/mimalloc-stats.h` exact stats | reachable only as `prof::stats().heap.*` | fully bound + `stats` safe module |
| `mi_purge_holes_report` | not declared | bound + `purge_holes_report()` |

38 fork C exports, 61 `mi_option_t` enumerators — all now bound, and checked on every build.

## Hazards checked (task step 4)

**(a) `mi_option_t` mirror.** There was no Rust mirror at all, so nothing was stale — but *adding* one creates the hazard, so it ships with its verifier. This fork inserts thirteen enumerators at indices **47..=59**, ahead of `_mi_option_last`; a mirror copied from upstream would compile, link, run, and set a *different* option than the caller named, forever, with no diagnostic. Two gates now:

- `tests/t19_layout.rs` compares every mirrored value against what the C compiler computed;
- `ci/check_rust_surface.py` compares the whole enumerator **sequence** against `include/mimalloc.h`. Sequence, not set: a mirror that merely dropped one enumerator still contains every correct *name*, and a set comparison would call it clean. `test_a_shifted_mirror_is_caught` is the negative control for exactly that.

**(b) Struct mirrors.** All six pre-existing mirrors (`mi_prof_stats_t`, `mi_prof_config_t`, `mi_prof_sample_info_t`, `mi_prof_module_info_t`, `mi_dhat_stats_t`, `mi_purge_holes_stats_t`) were audited field-by-field and were **correct — no mismatch found, no UB to fix**. They are now checked mechanically instead of by eye.

The mechanism: `rust/mimalloc-pprof/layout_probe.c` (a rust/-side verification artifact — not in the CMake build, not in the amalgamation, compiled by the crate's existing `cc::Build` so it always sees the library's own defines) publishes a name→value table of `sizeof`/`offsetof` for every mirrored struct plus the value of every mirrored enumerator and version constant. `tests/t19_layout.rs` looks each up **by name**: a field renamed or dropped on either side fails as a missing key, never as a coincidentally-equal number. The `mi_stats_t` entries are generated from the header's own `MI_STAT_FIELDS()` macro, so an upstream field addition fails the test without `layout_probe.c` changing — and a separate test asserts the C and Rust field *sets* are equal, so a mirror that is one field short cannot pass just because every offset it does check still matches.

**One correction to the task brief:** there are no idle-sweep gauges in `mi_stats_t`. It is byte-identical to upstream's. The gauges are in `mi_purge_holes_stats_t` (already bound as `MiPurgeHolesStats`), because the sweep also covers pages no heap owns and `mi_stats_t` cannot grow — it is embedded in a theap, at the meta-allocator's 8 KB block limit. Said so in the README table.

## API surface

The full three-column table is in [README.md → API surface](https://github.com/zackees/mimalloc-pprof/blob/feat/rust-api-parity/README.md#api-surface); condensed here.

### Sampled pprof profiler — `include/mimalloc/profile.h`

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_prof_start` | ✅ | `prof::start` |
| `mi_prof_start_seeded` | ✅ | `prof::start_seeded` |
| `mi_prof_start_ex`, `mi_prof_config_t` | ✅ | `enable_heap_profiling_with(&ProfConfig)` |
| `mi_prof_stop` | ✅ | `prof::stop` |
| `mi_prof_is_enabled` | ✅ | `prof::is_enabled` |
| `mi_prof_reset` | ✅ | `prof::reset` |
| `mi_prof_dump` | ✅ | `prof::dump_file` |
| `mi_prof_dump_writer` | ✅ | `prof::dump_to_vec` |
| `mi_prof_dump_proto` | ✅ | `prof::dump_proto_file` |
| `mi_prof_dump_proto_writer` | ✅ | `prof::dump_proto_to_vec` |
| `mi_prof_stats_get`, `mi_prof_stats_t` | ✅ | `prof::stats() -> ProfStats` |
| `mi_prof_snapshot_new` / `_visit` / `_free` | ✅ | `prof::samples() -> Vec<Sample>` |
| `mi_prof_modules_visit` | ✅ | `prof::modules() -> Vec<ModuleInfo>` |
| `mi_prof_visit` | ✅ | **sys only** — visitor runs under the profiler lock; an allocating Rust closure can deadlock |
| `mi_prof_debug_stats` | ✅ | **sys only** — deprecated |

### DHAT — `include/mimalloc/dhat.h`

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_dhat_start` / `_stop` / `_is_enabled` | ✅ | `dhat::start` / `stop` / `is_enabled` |
| `mi_dhat_stats_get`, `mi_dhat_stats_t` | ✅ | `dhat::stats() -> dhat::Stats` |
| `mi_dhat_dump` | ✅ | `dhat::dump_file` |

### Memory events — `include/mimalloc/memory-events.h` — **was entirely unbound**

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_memory_tracking_set_enabled` | ✅ **new** | `memory_events::set_enabled` **new** |
| `mi_memory_tracking_is_enabled` | ✅ **new** | `memory_events::is_enabled` **new** |
| `mi_memory_snapshot`, `mi_memory_snapshot_t` | ✅ **new** | `memory_events::snapshot()` **new** |
| `mi_memory_set_callbacks`, `mi_memory_callbacks_t`, `mi_memory_change_t` | ✅ **new** | `memory_events::set_callbacks(&'static Callbacks)` / `clear_callbacks` **new** |
| `mi_memory_visit_live_allocations` | ✅ **new** | `unsafe memory_events::visit_live_allocations` **new** |
| `mi_unwrapped_malloc` / `_free` / `_realloc` | ✅ | `unwrapped_malloc` / `unwrapped_free` / `unwrapped_realloc` |

### Exact allocator statistics — `include/mimalloc-stats.h` (upstream API)

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_stats_get`, `mi_stats_t` | ✅ **new** | `stats::get() -> Stats` **new** |
| `mi_stats_get_json` / `mi_stats_as_json` | ✅ **new** | `stats::json` / `Stats::to_json` **new** |
| `mi_stats_print_out` | ✅ **new** | `stats::print` **new** |
| `mi_stats_get_bin_size` | ✅ **new** | `stats::bin_size` **new** |
| `mi_subproc_stats_get` / `_get_exclusive` / `_get_json` / `_print_out` | ✅ **new** | `stats::subproc_get` / `subproc_get_exclusive` / `subproc_json` / `subproc_print` **new** |
| `mi_subproc_heap_stats_print_out` | ✅ **new** | `stats::subproc_heap_print` **new** |
| `mi_heap_stats_get` / `_get_json` / `_print_out`, `mi_heap_stats_merge_to_subproc` | ✅ **new** | **sys only** — take a `mi_heap_t*`, and v3 removed `mi_heap_get_default` |

### Heap dump

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_heap_dump_json` | ✅ | `heap_dump_json(include_blocks, hash_addresses)` |
| `mi_heap_get_seq` | ✅ | **sys only** — needs a `mi_heap_t*`; the seq numbers are already in `heap_dump_json`'s output |

### Thread idle, scavenger and hole purging — a [Bun feature](https://github.com/zackees/mimalloc-pprof#bun-features)

There is no single "hole punch" function: the feature is reached through the idle-sweep calls, the `purge_holes*` options, and the `mi_purge_holes_stats_t` gauges.

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_on_thread_idle` | ✅ | `on_thread_idle()` |
| `mi_on_thread_idle_start` / `_end` | ✅ | `park_while_idle() -> Option<IdlePark>` (ends on drop) |
| `mi_scavenger_stop` | ✅ | `scavenger_stop()` |
| `mi_purge_holes_stats_get`, `mi_purge_holes_stats_t` | ✅ | `purge_holes_stats()` |
| `mi_purge_holes_report` | ✅ **new** | `purge_holes_report()` **new** |

### Options — `mi_option_t` — **was entirely unbound**

| C | Rust FFI (`sys::`) | Rust safe wrapper |
|---|---|---|
| `mi_option_t`, 61 enumerators (13 fork, indices 47–59) | `sys::mi_option_*`, `sys::MI_OPTIONS_IN_ORDER` **new** | `options::Opt` **new** |
| `mi_option_get` / `_get_clamp` / `_get_size` | ✅ **new** | `options::get` / `get_clamp` / `get_size` **new** |
| `mi_option_set` / `_set_default` | ✅ **new** | `options::set` / `set_default` **new** |
| `mi_option_is_enabled` / `_enable` / `_disable` / `_set_enabled` / `_set_enabled_default` | ✅ **new** | same names under `options::` **new** |
| `mi_options_print_out` | ✅ **new** | `options::print` **new** |

Fork options with `Opt` constants: `PROF`, `PROF_SAMPLE_RATE`, `PROF_BT_MAX`, `PROF_ACCUM`, `PROF_SEED`, `PROF_MAX_BYTES`, `MEMORY_EVENTS`, `PURGE_ZEROES`, `SCAVENGER`, `PURGE_HOLES`, `PURGE_HOLES_EAGER_ZERO`, `PURGE_HOLES_MIN_INTERVAL`, `PURGE_HOLES_FULL_EVERY`.

## Design notes worth reviewing

- **`memory_events::set_callbacks` takes `&'static Callbacks`** of plain `fn` pointers, not closures. `memory-events.h` says the `arg` pointers are caller-owned and must stay valid until replaced; `'static` is the type system saying the same thing. One `extern "C"` trampoline serves all three slots (the kind is in the record) and catches unwinds, since a panic must not cross the C frame.
- **`mi_memory_change_kind_t` is mirrored as `c_int`, not a Rust `enum`.** The value comes from C, and materialising an out-of-range discriminant into a `#[repr(i32)]` enum is UB. `ChangeKind::from_raw` returns `Option` and the trampoline ignores an unknown kind.
- **`visit_live_allocations` is `unsafe`**, with the visitor's no-allocation / no-reentry contract in its `# Safety` block. `mi_heap_visit_blocks`'s own precondition (no concurrent frees into the walked heaps) is genuinely a soundness condition, not just a deadlock risk.
- **The print entry points return `()`, not `String`.** Routing mimalloc's output sink through a Rust closure would allocate from inside a callback the allocator drives — including, for `mi_purge_holes_report`, from inside a walk over the free lists being reported. `json()` is the capture route.
- **Nothing is feature-gated.** `src/memory-events.c`, the option table and the statistics compile into the C library in every configuration, so gating them on `pprof` would be wrong. All new tests run in both feature modes.
- **Test layout.** Memory-events tracking, the callback table and the running totals are process-global, and cargo runs one binary's tests on several threads — so a second test allocating on another thread lands in the counters the first is asserting deltas on. `t20` is therefore one test in sections, and the two tests with a real quiescence precondition (the `mi_heap_visit_blocks` walk; the read-only proof for `mi_purge_holes_report`, which the background scavenger would otherwise race) each get their own binary. This was found empirically — both flaked before being split.

## Gate: `ci/check_rust_surface.py`

Parses the C headers and `sys.rs`/`lib.rs`; fails when a fork export has no `sys.rs` declaration, when a bound export has no `lib.rs` call site, or when the option sequence diverges. Allowlists carry a reason each and a **stale entry is itself a failure** — an allowlist entry that no longer describes anything real is a claim nobody checked. Three entries today, all sys-only-by-design.

Deliberately **no `git show 6def7be9`** to separate fork additions from upstream: `actions/checkout@v4` is depth 1 and this fork's history is unrelated to upstream's, so that object does not exist in CI. The eight fork additions inside the two shared headers are a literal list — and every name on it is verified to still be an export, so a rename cannot quietly empty it.

Wired into `python-lint.yml` (with `--selftest`) and `rust-native.yml`, plus `ci/verify_local.py`'s `lint` and `rust` configs. `rust-native.yml`'s path filter gains `include/**`: without it a C-only PR could shift the option enum and no Rust job would look.

## Verification

```
cargo test -p mimalloc-pprof                       # green x9 (25 binaries + 6 doctests)
cargo test -p mimalloc-pprof --no-default-features # green (11 binaries + 6 doctests)
cargo clippy -p mimalloc-pprof --all-targets       # clean
cargo run -p xtask -- check                        # vendored amalgamation matches src/include/
cargo publish --dry-run -p mimalloc-pprof --locked # green
python ci/check_crate_package.py target/package    # verified publish archive
cargo package -p mimalloc-pprof --list             # layout_probe.c ships with the crate

uv run ci/check_rust_surface.py
  -> 38 fork exports bound, 61 mi_option_t enumerators mirrored in order
uv run ci/check_rust_surface.py --selftest         -> parsers OK
uv run ci/check_doc_snippets.py                    -> PASS: 9 blocks checked, 3 skipped
uv run --with ruff==0.12.10 ruff check ci          -> All checks passed
uv run --with ruff==0.12.10 ruff format --check ci -> 43 files already formatted
uv run --with pyright[nodejs]==1.1.411 pyright     -> 0 errors, 0 warnings
uv run --with pytest==8.3.4 --with pyyaml==6.0.2 -m pytest ci/tests -q  -> 298 passed

uv run ci/verify_local.py --only rust,lint
  rust | PASS | 103.1s
  lint | PASS |  49.0s
```

No C-core paths touched, so no amalgamation re-run and no `release`/`off`/`debug-full` configs needed. Commits are split `feat(rust):` / `ci:` / `docs:` per the repo rules; version deliberately not bumped (this repo bumps in a dedicated release PR), and a `## Unreleased` CHANGELOG section was added instead.

## Post-review fixes (in this PR)

- `ci:` **the wrapper check was substring-matched**, and C's names nest — `sys::mi_prof_start_seeded` contains `sys::mi_prof_start`, `sys::mi_option_get_clamp` contains `sys::mi_option_get`. An unwrapped function would have been reported as wrapped on the strength of its longer sibling's call site, i.e. exactly the regression the gate exists to catch. Now word-anchored, with the negative control in both the pytest suite and the script's `--selftest`. Nothing in the tree was actually unwrapped, so this closed a hole in the gate, not in the bindings.
- `fix(rust):` `purge_holes_report`'s doc comment invented a sink argument. `mi_purge_holes_report(void)` has none; the sink is process-wide (`mi_register_output`, unbound).
- Also verified the two rust-native steps `verify_local` does not cover: `cargo publish --dry-run -p mimalloc-pprof --locked` and `ci/check_crate_package.py target/package` — both green, and `layout_probe.c` ships in the archive.
- `## Unreleased` is new to this CHANGELOG; nothing in `.github/workflows`, `ci/` or `rust/xtask` parses that file, so no release tooling is affected.

## Second review round (applied)

- `ci:` **a doc comment naming a C function is not a call to it.** `lib.rs` documents each wrapper by naming the function it wraps, so the raw-text search would have kept reporting a *deleted* wrapper as wrapped. Comments are stripped first now — block, whole-line `//`/`///`/`//!`, and trailing `//` — with the trailing rule refusing to fire after a colon so `https://` in a doc link does not swallow the rest of its line. Both halves have fixtures, in `--selftest` and in ci/tests. The full check still passes afterwards, which is the useful part: every wrapped export has a real call site, not merely a mention.
- `fix(rust):` **`visit_live_allocations`' `# Safety` now forbids panicking.** Raising a panic allocates its payload and message through the global allocator, reentering mimalloc mid-walk — exactly what the no-allocation bullet forbids. `catch_unwind` stops the unwind crossing the C frame; it does not, and cannot, prevent that allocation, which has already happened by the time it runs.
- `fix(rust):` `t20`'s unwrapped-allocation section calls `scavenger_stop()` before its strict-equality window, the way `t23` already does — the background scavenger runs on a timer and could allocate inside it. Verified over 5 consecutive runs of t20/t21/t23 and 9 of the full suite.
- `docs:` the gate requires an unbound **fork** export to fail; upstream's exports are bound opportunistically. README no longer promises the stronger guarantee.

Follow-up filed as an issue rather than fixed here: `layout_probe.c` gates struct *layouts* but not function *signatures*.

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -41,6 +41,10 @@ jobs:
         run: |
           python3 ci/check_isa_baseline.py --selftest
           python3 ci/check_internal_state.py --selftest
+          # Parses C headers and Rust source with regexes; a regex that matches nothing
+          # reports "clean" exactly as loudly as one that works.
+          python3 ci/check_rust_surface.py --selftest
+          python3 ci/check_rust_surface.py
           # A gate script that imports cleanly but crashes on argv is still broken.
           python3 ci/memory_gate.py 2>&1 | grep -q "Exit codes"
           python3 ci/check_isa_baseline.py --help > /dev/null
--- a/.github/workflows/rust-native.yml
+++ b/.github/workflows/rust-native.yml
@@ -7,6 +7,12 @@ on:
     paths:
       - 'rust/**'
       - 'ci/check_crate_package.py'
+      - 'ci/check_rust_surface.py'
+      # The C headers ARE an input to this job: ci/check_rust_surface.py fails when a
+      # header grows an export or an mi_option_t enumerator that rust/mimalloc-pprof's
+      # hand-written bindings do not mirror. Without these paths a C-only PR could add
+      # an API (or shift the option enum) and this job would never look.
+      - 'include/**'
       - '.github/workflows/rust-native.yml'
 
 concurrency:
@@ -48,6 +54,14 @@ jobs:
       - name: xtask check (vendored C amalgamation must match src/include/)
         if: runner.os != 'Windows'
         run: soldr cargo run -p xtask -- check
+      # The bindings in src/sys.rs are hand written, so nothing in the Rust build notices
+      # when C grows an export or shifts mi_option_t. This is what notices. Run before
+      # `cargo test` so a missing binding is reported as a missing binding rather than as
+      # a compile error three steps later.
+      - name: Rust binding surface covers the fork's C API
+        if: runner.os != 'Windows'
+        shell: bash
+        run: python ../ci/check_rust_surface.py
       - run: soldr cargo test --locked
       - name: cargo publish --dry-run (mimalloc-pprof)
         if: runner.os != 'Windows'
--- a/README.md
+++ b/README.md
@@ -332,8 +332,9 @@ For a measured chart of what this buys on a churn workload, see
 
 Every API this fork adds, in all three places it can be reached from: the C header,
 the crate's raw FFI module (`mimalloc_pprof::sys`, `unsafe`), and the crate's safe
-wrapper. There are no gaps — `ci/check_rust_surface.py` fails the build if a C export
-or an `mi_option_t` enumerator appears without a binding, and
+wrapper. There are no gaps — `ci/check_rust_surface.py` fails the build if a **fork** C export
+or an `mi_option_t` enumerator appears without a binding (upstream's own exports are
+bound opportunistically, not by requirement), and
 `rust/mimalloc-pprof/tests/t19_layout.rs` checks every mirrored struct layout and
 option value against what the C compiler actually laid out.
 
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ Three instruments, each shown in Rust and C: **pprof** sampled profiling for
 production, **exact allocator stats** to check a sampled profile against, and
 **DHAT** exact profiling for focused investigations.
 
+In a hurry, or looking for one specific call? [**API surface**](#api-surface) below
+is the whole thing in one table — every C entry point, its raw Rust FFI declaration,
+and its safe Rust wrapper.
+
 ### pprof: sampled heap profiling
 
 #### Rust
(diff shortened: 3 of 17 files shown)


zackees:
Follow-up from this review filed as #322 (layout_probe.c gates struct layouts but not function signatures — redeclare each bound prototype so an incompatible C signature change is a compile error, C11 6.7p4 / MSVC C2373). Deliberately not fixed here.

PR #333, opened 2026-09-03
fix(bench-harness): tolerate a child that exits before reading its request (#323)

### LINKED ISSUE #323: flake(rust-native): bench-harness rejections::benchmark_rejects_partial_or_extra_child_output fails intermittently on ubuntu-latest
Seen on PR #321 (which does not touch `rust/bench-harness`), run 33799764470, job `test (ubuntu-latest)`:

```
test benchmark_rejects_partial_or_extra_child_output ... FAILED
thread 'benchmark_rejects_partial_or_extra_child_output' panicked at bench-harness/tests/rejections.rs:47:5:
assertion failed: matches!(run_benchmark(config(), &child(Some("partial"))), Err(BenchError::InvalidChildResponse(_)))
test result: FAILED. 6 passed; 1 failed
```

The test spawns `stress-child` with `BENCH_HARNESS_TEST_RESPONSE=partial` and expects `Err(InvalidChildResponse)`. The panic message does not show which variant came back, so the first fix is diagnostic: replace the `matches!` asserts with a match that prints the actual `Result` on mismatch (or `assert!(matches!(..), "got {:?}", r)`), so the next occurrence names the variant (likely a spawn/io/timeout error under runner load rather than a parse result). Then decide whether the test needs a retry-free hardening (e.g. a longer child timeout in `config()`).

Re-run of the same head is in progress; result to be appended.

zackees:
Closes #323 (seen twice today on unrelated PRs #321 and #330, on different assertions of the same test).

**Cause.** `run_child_sample` spawns the child, then writes the JSON request to its stdin and returns `BenchError::WriteRequest` on any write error. The rejection tests' child (`BENCH_HARNESS_TEST_RESPONSE=partial|extra|nonfinite`) prints its bogus response and exits without reading stdin. When the child wins the race the parent's write hits `EPIPE`, so the harness reports `WriteRequest(Broken pipe)` instead of the `InvalidChildResponse` the test expects. Same thing would misreport a real child crash-before-read.

**Fix.** Treat `BrokenPipe` on the request write as "the child has already exited" and fall through to the existing wait/parse path, so the exit status and output determine the error regardless of ordering. Other write errors still return `WriteRequest`. The three rejection assertions now print the actual `Result` on mismatch (the diagnostic step from #323).

**Verification.** `cargo test -p bench-harness` green; the rejections test run 6× consecutively, all green; clippy and fmt clean on the crate.

### DIFF
--- a/rust/bench-harness/src/lib.rs
+++ b/rust/bench-harness/src/lib.rs
@@ -421,12 +421,20 @@ fn run_child_sample(
         .stderr(Stdio::piped())
         .spawn()
         .map_err(BenchError::Spawn)?;
-    process
-        .stdin
-        .take()
-        .expect("child stdin piped")
-        .write_all(&request)
-        .map_err(BenchError::WriteRequest)?;
+    {
+        let mut stdin = process.stdin.take().expect("child stdin piped");
+        // A child that exits before reading its request (a crash, or a deliberately
+        // malformed test child that prints and returns) closes the pipe under us. That
+        // is not a request-delivery failure: fall through and let the child's exit
+        // status and output say what went wrong, so the reported error is the same
+        // whichever side won the race (#323).
+        match stdin.write_all(&request) {
+            Ok(()) => {}
+            Err(error) if error.kind() == std::io::ErrorKind::BrokenPipe => {}
+            Err(error) => return Err(BenchError::WriteRequest(error)),
+        }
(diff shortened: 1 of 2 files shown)


zackees:
Superseded: the identical BrokenPipe fall-through landed as 9849fbad on PR #330's branch (another agent hit the same flake, 7/30 locally) and merges with #330. #323 stays linked to #330.

PR #334, opened 2026-09-03
bench: hole-purging chart under jemalloc, upstream mimalloc, Bun mimalloc and mimalloc-pprof (Performance section)

zackees:
Moves the hole-purging chart out of "Bun features" and into `## Performance`, as the
last subsection after the existing benchmark subsections — and, while it was being
moved, gives it something to be compared against. Until now it could only answer an
in-tree A/B (`MIMALLOC_PURGE_HOLES=0` vs `=1` in one binary), because no competitor
allocator has that option to switch off.

The new `ci/bench_hole_purging_allocators.py` runs the **same churn workload** —
150k × 512 B + 100k × 1 KiB + 50k × 2 KiB blocks, a scattered 1-in-20 kept alive,
the rest freed, then a 10 s idle window — once per allocator, each linked against the
static library `rust/benchmark-suite/allocators/allocator-lock.json` builds.

## Measured

AMD Ryzen 7 3700X (16 threads), Linux 6.18.48, `taskset -c 0-3`, at commit
`86e0c5a4` on this branch. Best of 3 runs by after-idle RSS, the same rule for every
allocator. Peak is the kernel's `VmHWM`.

| series | what runs during idle | peak RSS | after 10 s idle | returned |
|---|---|---|---|---|
| mimalloc-pprof | `mi_on_thread_idle()` every 100 ms | 278.7 MB | 71.6 MB | **74 %** |
| jemalloc | nothing (a plain idle process) | 282.5 MB | 282.5 MB | **0 %** |
| jemalloc + explicit purge | `mallctl("arena.<all>.purge")` every 100 ms | 282.4 MB | 72.6 MB | 74 % |
| Bun mimalloc | `mi_on_thread_idle()` every 100 ms | 278.7 MB | 71.7 MB | **74 %** |
| upstream mimalloc | nothing (a plain idle process) | 276.7 MB | 276.7 MB | **0 %** |

Diagnostics — measured, recorded in the report JSON, shown below a rule in the table
SVG, never charted:

| series | what runs during idle | peak RSS | after idle | returned |
|---|---|---|---|---|
| mimalloc-pprof, no idle hook | nothing | 278.8 MB | 278.8 MB | 0 % |
| upstream mimalloc, `mi_collect(false)` | `mi_collect(false)` every 100 ms | 276.6 MB | 227.8 MB | 18 % |
| upstream mimalloc, `mi_collect(true)` | `mi_collect(true)` every 100 ms | 276.6 MB | 227.8 MB | 18 % |
| jemalloc, `background_thread:true` | nothing; `JE_MALLOC_CONF=background_thread:true` | 282.6 MB | 282.6 MB | 0 % |
| jemalloc, `background_thread:true`, 30 s | the same, 30 s window | 282.6 MB | 72.7 MB | 74 % |

Every series in the committed run is 3-of-3 with a run-to-run spread of at most
0.1 MB; all three runs of each are in `.github/assets/allocator-idle-report.json`.
The one row that is genuinely borderline is jemalloc's `background_thread:true` at
10 s — see below.

## What this does and does not support

**The finding that was not anticipated, first: `mimalloc-pprof` with no idle hook
returns 0 %.** On this workload a survivor lands every 20 blocks, so essentially
every 64 KiB mimalloc page still holds something alive and there is no fully-free
arena memory for the background scavenger to return by itself. The entire 74 % is
hole purging, and it happens on the `mi_on_thread_idle()` call an embedder makes
before it blocks. The README says so in the same breath as the 74 %, the chart SVG
carries it in its own caption, and a bridging sentence was added to the "Bun
features" prose (which says an idle process "stops sitting on memory it no longer
needs") so the two cannot be read as contradicting.

**jemalloc:** the owner's phrasing holds — *jemalloc can return memory if you ask; it
doesn't after idle* — but only for the default configuration, and the PR/README say
exactly that. Its decay is advanced by allocation activity, not by time spent idle,
so the default line is flat; an explicit `mallctl("arena.<all>.purge")` gets 74 %.
Its opt-in `background_thread:true` sits **exactly on jemalloc's own boundary**: the
default `dirty_decay_ms` is 10 s and this window is 10 s, so the committed run
returns 0 % inside 10 s and 74 % over 30 s — and an earlier session of the same
benchmark had that row returning in two runs of three at 10 s. Both windows are on
the chart caption and in the diagnostics for that reason. Not claimed anywhere: that
jemalloc's mechanism is worse — the cost of either hook was not measured.

**upstream mimalloc is the hole-punch comparison, and it is the strongest honest
sentence here:** no `mi_on_thread_idle` at all, and **18 %** when given
`mi_collect(false)` on the same tick — and the same **18 %** when given
`mi_collect(true)`, which forces the purge rather than honouring upstream's 1000 ms
`purge_delay`. That is a committed diagnostic row precisely so the obvious objection
("you gave it the non-forcing collect") has an answer in the artifact set rather than
in a comment. Page granularity is the ceiling, not the delay. It also independently
reproduces the 18 % the off-vs-on chart measures for this fork's own scavenger with
hole purging disabled. Whole-page return gets 18 %; punching holes in still-used
pages gets 74 %. Caveat repeated where the number is claimed: the upstream row is
`bcee5a88` (v3.4.3) and not this fork's base, so upstream-vs-fork mixes base and fork
changes; Bun's row is the like-for-like line and lands on the same 74 %.

*Not measured, offered only as an explanation:* jemalloc's per-size-class slabs are
much smaller than mimalloc's 64 KiB pages, so a 1-in-20 survivor rate leaves many
jemalloc slabs completely free and a forced purge can return them without needing
anything hole-shaped. That is a structural difference, not something this benchmark
established.

## Method, and the fairness rules that are in the driver rather than the prose

- **Nothing in the harness allocates.** The block table is `mmap`ed, RSS is read with
  `open`/`read`/`close` into a stack buffer, samples leave through `write(2)`.
  `libmimalloc.a` carries `alloc-override.c`, so in the three mimalloc children every
  libc `malloc` in the process is mimalloc's, while in the `je_`-prefixed jemalloc
  children it is glibc's — a `printf` per idle tick would be allocator traffic in one
  arm and not the other, and allocator traffic during "idle" is exactly what
  jemalloc's decay needs in order to fire.
- **Peak is `VmHWM`,** read at the end, floored at the first sample. Sampling starts
  at the top of the idle window, so `max(sample)` would be post-free RSS; and both
  counters come from per-CPU caches flushed periodically, so `VmHWM` can read *below*
  a `VmRSS` sample taken moments earlier (it did in 13 of 27 runs). Both are good to
  a few hundred kB, which the table says.
- **`mi_on_thread_idle` support is read out of each allocator's extracted header,**
  not assumed — upstream at its pin has none, which the script detects.
- **A knob that never turned on is fatal.** A diagnostic that tunes an allocator
  reads the setting back out of it and refuses to publish on a mismatch. This caught
  a real bug mid-development: the first `background_thread` row set `MALLOC_CONF`,
  which a `je_`-prefixed jemalloc ignores in complete silence (it reads
  `JE_MALLOC_CONF`), and it produced a perfectly plausible 0 % that would have
  shipped as "even jemalloc's decay thread returns nothing". `ci/tests` has the
  RED/GREEN for it.
- **A series whose repetitions disagree says so on the image.** The charted figure is
  the best of 3, so a series that returned memory in only some of its runs would be
  represented by its lucky run alone. The table renders an "only N of M runs"
  footnote for any such series, counted from the run records, and the chart caption
  carries the same count — so it cannot go stale on the next measurement.
- Source acquisition, checksums and build commands are reused from
  `ci/build_benchmark_allocators.py`, so the pins are the locked ones.
  `build_records` is deliberately not called — it also cargo-builds a
  `benchmark-child` per allocator that nothing here would run.

## Caveats

- **tcmalloc is not in this chart.** Its locked build needs bazel, which the
  measuring machine did not have; a substitute build would not be the pinned one.
  Named in the chart's prose, in `docs/benchmarks.md`, and in the report JSON's
  `skipped_allocators`.
- **Both chart pairs are dev-box measurements, not runner output** — the same machine
  in both cases, but two separate runs with different selection rules (median of 3
  for the off-vs-on pair, best of 3 for the cross-allocator pair). Each SVG names its
  own machine, kernel, commit and rule; the README and `docs/benchmarks.md` say it in
  words as well, and the Performance intro's "GitHub-hosted, no self-hosted hardware"
  promise is now scoped to the throughput matrix so the page does not falsify itself
  two screens later. The off-vs-on pair is committed unchanged.
- Downloading the pinned archives needed `SSL_CERT_FILE` pointed at the system CA
  bundle in this sandbox. That is an environment quirk, not a script change.

## Placement and anchors

`### Memory returned after idle` is the last subsection of `## Performance` and holds
both chart pairs — the four-allocator one, then the moved off-vs-on pair under
`#### Isolating this fork's own contribution`. "Bun features → Hole purging,
measured" keeps every word of its prose and gains a one-line pointer. Anchor updates:
the contents list, the At-a-glance "hole punch" bullet, the pointer under the
scavenger API section, and the feature table row. `docs/benchmarks.md`'s "Not covered
by the five-allocator matrix" section claimed a Bun series was possible but had been
left for a run on the reference machine — this PR makes that false, so it is replaced
with a full methodology section for both scripts.

## Verification

`ci/tests` (334 tests) green, including `test_bench_hole_purging_allocators.py`,
which re-renders **all eight** committed SVGs (two scripts × chart/table ×
light/dark) from the committed CSV/JSON and fails on any difference, asserts the
README `alt` text is the same sentence the SVG carries, guards the table's column
widths (an SVG has no layout engine — the `background_thread` rows overflowed into
the peak column once already, and only a rendered PNG showed it), and covers the
run-disagreement footnote synthetically since the committed run is 3-of-3
everywhere. `--check` for both scripts in both modes is now a step in
`python-lint.yml` and a command in `ci/verify_local.py`'s lint config; neither
re-measures. Also green: `ci/check_doc_snippets.py`, ruff 0.12.10 + pyright 1.1.411
strict, and `ci/verify_local.py --only lint`.

Every figure in the charts, the table SVGs and the README comes from the run
described above.

Was stacked on #330 during development; #330 has since merged, so this is rebased
onto `main` and targets `main` directly.


zackees:
Follow-up in 86e0c5a4, no measured number changed:

- The section's lead-in had its own logic backwards (the first two bullets are the caveats; upstream's 18 % is the claim that survives them).
- The Performance intro's trust sentences ("Every result is GitHub-hosted… no self-hosted hardware", "Charts below are regenerated by the scheduled dashboard run") now scope themselves to the throughput matrix, and name the memory charts as the labelled exception — otherwise the page falsifies its own claim two screens later.
- The At-a-glance "hole punch" bullet now carries the competitor numbers it could not carry before (0 % for upstream mimalloc and for default-config jemalloc).

And one check a reviewer would otherwise have to ask for. The upstream diagnostic uses `mi_collect(false)`, so the obvious objection is "you handed upstream the non-forcing collect and it just lost to `purge_delay` (1000 ms upstream)". Re-running the same driver against the same pinned upstream build with `mi_collect(true)` — which forces the purge instead of honouring the delay — returns **17.6 % on all three runs**, identical to the charted `mi_collect(false)` figure. Page granularity is the ceiling there, not the delay. Noted in the README bullet.

zackees:
Review addressed in 857bc43f (`ci:`) and ffd83885 (`docs:`), no force-push. Assets are a full re-run at `86e0c5a4` on this branch.

**B1 — `mi_collect(true)` had no committed backing.** Took option (a): new `BENCH_IDLE_MI_COLLECT_FORCE` mechanism and an `upstream-mimalloc-collect-force` diagnostic series. It measures **17.6 % on all three runs**, identical to `mi_collect(false)`, and is now a row in the table SVG and the report JSON. The README cites it and says why the row exists.

**B2 — the `background_thread` 10 s row and the false "≤ 0.2 MB" claim.** Fixed at the root rather than in prose: the table renders an "only N of M runs" footnote for any series whose repetitions disagree, and the chart caption carries the same count — both computed from the run records, so neither can go stale. The caption now also always states *both* jemalloc background windows, because that row sits exactly on jemalloc's default `dirty_decay_ms` (10 s = this window) and has legitimately landed on both sides: the re-run came back **0 % in 3 of 3 at 10 s and 74 % at 30 s**, where the earlier session was 73 % in 2 of 3. README and PR body say that, including the earlier session, explicitly. In the committed run every series is 3-of-3 with spread ≤ 0.1 MB, so no footnote renders; a synthetic test covers that path.

**B3 — "not the reference machine".** Correct, that was false: both pairs are the same dev box. The PR body now says both are dev-box measurements differing in selection rule (median-of-3 vs best-of-3), and the same statement is in the README (C7) and `docs/benchmarks.md` (C8).

Should-fixes, all applied: **C1** fifth+sixth caption lines so the cooperative caveat travels with the image; **C2** at-a-glance now reads "18 % if you call `mi_collect` yourself"; **C3** all eight SVGs (both scripts × chart/table × light/dark) re-rendered by `ci/tests`, and `--check` for both scripts in both modes wired into `python-lint.yml` **and** `ci/verify_local.py`'s lint config, with `docs/benchmarks.md` describing what is actually enforced; **C4** repetition count threaded through the caption and the selection rule; **C5** the overlap note computes its own spread (now "within 0.9 MB"); **C6** `peak_kb = max(peak_kb, samples[0].rss_kb)` plus a note that both counters are good to a few hundred kB; **C7/C8/C9** as above, plus the "not one base" caveat repeated where the 18 % is claimed.

One thing the review did not ask for but the C3 work exposed: the Performance intro promised "Every result is GitHub-hosted… no self-hosted hardware" two screens above a dev-box subsection. That is now scoped to the throughput matrix, with the memory charts named as the labelled exception.

Green: 334 pytest, both scripts' `--check` in both modes, `check_doc_snippets.py`, ruff 0.12.10, pyright 1.1.411 strict, `verify_local.py --only lint`.

zackees:
Two more, pushed as 6db49204 (`ci:`) and 579d7602 (`docs:`):

- The run-disagreement footnote is the one bit of table layout no committed asset exercises — every series in the committed run is 3-of-3, so the rendered table always has exactly one footnote and the byte-for-byte comparisons never see the two-line shape. The synthetic flaky-series test now parses its own SVG and asserts exactly two 11 px footnote lines, ≥12 px apart, with the last inside the viewBox. That failure mode (silent overprint) already bit the "what runs during idle" column once.
- Removed the README's "an earlier session had it returning in two runs of three at 10 s" parenthetical. That session's report JSON is not in the repo, so it is a figure a reader cannot check in `.github/assets/allocator-idle-report.json` — it does not belong in the README even hedged. The structural reason the row is borderline (default `dirty_decay_ms` = 10 s = the window) stands on its own and is what makes the committed 0 %/74 % pair honest. The cross-session history stays in the PR description, labelled as development history rather than as a measurement — so the README and the body deliberately differ on that one point.

PR #393, opened 2026-09-07
docs: lead the README with the feature matrix and the scaling charts

zackees:
Owner request, four parts.

### 1. Feature comparison first

`## Feature comparison` now sits directly under the intro, before **Contents**. It is
the one element that answers "what does this fork give me that the others don't"
without reading prose.

### 2. Then the thread-scaling charts

`## Thread scaling by allocation pattern` follows it, so anyone landing here sees the
performance characteristics in the same screen as the feature claims. Both keep their
contents-list entries, marked *(above)*.

The pattern table that describes the four workloads stayed behind in **Performance**
when its heading moved up; it gets its own heading (`### What the four scaling patterns
stress`) and links back to the charts. The Performance prose that said "the throughput
charts below" now points up.

### 3. Left-justified cells in the image

`cell_svg` centred the glyph-plus-note pair inside its column, so each glyph was
displaced by half the length of its own note. Measured before the fix: the marks in one
column landed on **17–20 distinct x-positions, spanning up to 88px**. A reader scanning
a column for crosses had to read every cell to find them.

Cells now start at a fixed inset (`column_left(index) = COL_START + COL_W*index + 12`),
so every glyph in a column sits on **exactly one x**. Column headers moved with them. A
glyphless *value* cell (the RSS row) uses the note x rather than the glyph x, so its
text lines up with the notes above and below instead of jutting into the gutter.

### 4. Markdown matrix into a `<details>`

Directly under the image the text matrix is redundant for a human, but it is what a
screen reader, ctrl-F and an LLM reading this file actually consume — so it stays in the
file, one fold away, under *"The same matrix as text"*.

### Verification

- `ci/tests/test_render_feature_table.py` — 33 passed
- `uv run python ci/render_feature_table.py --check` — `up to date: README.md region,
  allocator-features-light.svg, allocator-features-dark.svg`
- ruff check / ruff format / pyright — clean
- Both SVGs rasterized and inspected; alignment confirmed in light and dark

Docs and `ci/` only — no C-core and no `rust/` paths.

https://claude.ai/code/session_01UfNcj3bdhvjVF9xPPXKPAA

### DIFF
--- a/.github/assets/allocator-features-dark.svg
+++ b/.github/assets/allocator-features-dark.svg
@@ -6,571 +6,571 @@
 <text x="24" y="26" font-size="17" font-weight="600" fill="#e6edf3">Allocator feature comparison</text>
 <text x="24" y="46" font-size="11" fill="#8b949e">Sourced cell by cell in docs/allocator-features.json. Upstream and Bun are the exact commits this fork pins to and ports from; jemalloc is the build in the benchmark lockfile. The measured row is the committed churn benchmark,</text>
 <text x="24" y="61" font-size="11" fill="#8b949e">whose upstream arm is v3.4.3 (bcee5a88).</text>
-<text x="525.0" y="93" font-size="12.5" font-weight="700" text-anchor="middle" fill="#e6edf3">mimalloc-pprof</text>
-<text x="525.0" y="108" font-size="10" text-anchor="middle" fill="#8b949e">this fork, 0.9.x</text>
-<text x="735.0" y="93" font-size="12.5" font-weight="600" text-anchor="middle" fill="#8b949e">Microsoft MiMalloc-V3</text>
-<text x="735.0" y="108" font-size="10" text-anchor="middle" fill="#8b949e">v3 dev3 @ 6def7be9</text>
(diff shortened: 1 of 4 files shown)


zackees:
CI: 48 green, 2 skipped, and `run-windows` red on `windows-gnu-x64-gated::test-purge-all-no-scavenger`.

That failure is **pre-existing on `main`**, not from this PR — this diff is README + `ci/render_feature_table.py` only, no C. Evidence:

- `main` @ `e8c4a408` ([run 34089530368](https://github.com/zackees/mimalloc-pprof/actions/runs/34089530368)) fails the same `windows-gnu-x64-gated::test-purge-all-no-scavenger`.
- `main` @ `abf27abf` ([run 34085581755](https://github.com/zackees/mimalloc-pprof/actions/runs/34085581755)) fails `windows-gnu-x64-release::test-purge-all`.
- Re-running this PR's job reproduced it with a *different* assertion set (first run: the C2 series, rerun: G6), which is the signature of the flaky/broken owner-gate purge path, not of anything in this diff.

Merging with admin; the win-gnu purge-all failure is tracked separately as the remaining 0.11.0 release blocker.

PR #398, opened 2026-09-13
fix: make heap diagnostics ownership-safe without hot-path changes

### LINKED ISSUE #374: Two diagnostic walks race a sweeper/owner: use-after-poison in the holes reporter (ASan, owner-gate) and a _mi_ptr_page assertion in mi_heap_dump_json (guarded)
## Summary

Two read-only diagnostic walks intermittently read heap memory that another thread is
changing underneath them. Both fired once each on PR #372 (`c-unit` `run-linux` and
`asan.yml`'s owner-gate row) and both went green on a rerun of the *identical* commits, so
they present as flakes. Both are real races; neither is caused by #372 (both failing
configurations build with `MI_DHAT` defaulted ON, i.e. the same code as `main`).

- **A** is a #366 gap: `_mi_purge_holes_report_collect` never enters the owner gate, so in a
  gated build the reporting thread is PARKED while it walks its own free lists, and the
  scavenger's claimed sweep rewrites those lists under it. `tld->theaps_lock` protects the
  reporter from the sweep's *hole* phase only, not from its *collect* phase.
- **B** is the page-retirement race that `mi_heap_visit_blocks` is already documented not to
  be safe against, reaching a debug assertion. `MI_GUARDED` with
  `MIMALLOC_GUARDED_SAMPLE_RATE=1` is what makes the test's pages empty and get freed during
  the concurrent dump; the test's own comment assumes they never do.

## A. `test-purge-holes` — use-after-poison walking a free list (ASan, owner-gate)

`address sanitizer (clang, Debug, owner-gate)`, run
[34055444891 attempt 1](https://github.com/zackees/mimalloc-pprof/actions/runs/34055444891):

```
==40403==ERROR: AddressSanitizer: use-after-poison on address 0x7fc86ff50000
    #0 mi_block_nextx                 include/mimalloc/internal.h:1772:50
    #1 mi_block_next                  include/mimalloc/internal.h:1794:22
    #2 mi_holes_mark_free_list         src/page-holes.c:933:9
    #3 _mi_page_holes_report_page      src/page-holes.c:998:3
    #4 mi_theap_page_holes_report      src/page-holes.c:1223:3
    #5 _mi_theap_visit_pages           src/theap.c:47:12
    #6 _mi_purge_holes_report_collect  src/page-holes.c:1242:7
    #7 test_report_is_read_only        test/test-purge-holes.c:1155:3
    #8 main                            test/test-purge-holes.c:1823:3
SUMMARY: AddressSanitizer: use-after-poison in mi_block_nextx
```

The other three ASan rows (clang Debug ungated, clang RelWithDebInfo, gcc RelWithDebInfo) ran
`test-purge-holes` in the same attempt and passed. Only the **owner-gate** row failed.

**Why this is a race and not ASan noise.** `mi_block_nextx` unpoisons the 16-byte block
header, reads `next`, and re-poisons (`internal.h:1769-1777`). A poisoned free block is the
normal state; for the read on line 1772 to trip, the header's shadow must be poisoned again
*between* the unpoison on 1769 and the read — by another thread. Single-threaded this cannot
fault.

**What the faulting address is.** By construction of the stack, `0x7fc86ff50000` is the
free-list block whose header `mi_block_nextx` is reading. Its 64 KiB alignment is ordinary:
with separated page meta (`src/arena.c:895-915`) `block_start == 0` for
`block_size > mi_good_size(MI_SMALL_SIZE_MAX)`, so block 0 of such a page (the
survivors-8192/16384/65536 pages left over from earlier cases in `main`) sits on a slice
boundary. It is *not* evidence of an OS-page-granular discard.

**Who re-poisons.** Not a hole discard: `_mi_os_discard` (`src/os.c:671-699`) never touches
the ASan shadow and skips even its eager `memzero` under `MI_TRACK_ENABLED`. The only code
that poisons exactly one block header is `mi_track_mem_noaccess(block, sizeof(mi_block_t))`
in `mi_block_nextx` / `mi_block_set_nextx` (`internal.h:1777,1789`) — i.e. a second thread
walking or rewriting the same free list, as `mi_page_thread_free_collect` and
`mi_page_free_collect_ex` do (`src/page.c:182,199,259,262`).

**Who that second thread is.** In a gated build every `MI_GATE_LEAVE` stores
`MI_PARK_PARKED` (`include/mimalloc/owner-gate.h`, `_mi_gate_leave`).
`_mi_purge_holes_report_collect` (`src/page-holes.c:1227-1257`) has no `MI_GATE_ENTER`, so
the test thread walks its own theaps while PARKED. The scavenger — started when the test's
earlier cases ran a second thread (`src/init.c:448`; `test_abandoned`,
`test_report_read_only_abandoned`) — claims any PARKED tld that is due
(`src/scavenger.c:243-254`), and the test sets `purge_holes_min_interval=0`
(`test-purge-holes.c:1772`), so every PARKED tld is always due. The claimed sweep runs
`_mi_thread_idle_work_ex` (`src/scavenger.c:120-140`) in two phases:

1. `_mi_theap_collect_foreign` → `mi_theap_collect_ex` (`src/theap.c:152-170`) →
   `_mi_theap_visit_pages(..., mi_theap_page_collect, ...)`: rewrites `page->free` /
   `page->local_free` (the poisoning writes above) and frees all-free pages. **Holds no
   lock.**
2. `_mi_purge_holes_of` (`src/page-holes.c:838-897`): **holds `tld->theaps_lock`** for the
   whole hole phase, including the abandoned pass (`page-holes.c:874-880` and its comment).

The reporter holds `tld->theaps_lock` too (`page-holes.c:1239`), so phase 2 cannot overlap
it — but phase 1 can, and phase 1 is what touches the free lists the reporter walks.

**Why only the gated row.** Ungated, a thread is PARKED only through
`mi_on_thread_idle_start`; the test's `mi_on_thread_idle()` is enter + work + leave on the
owner (`src/scavenger.c:803-813`), so the reporting thread is RUNNING and nothing can claim
it. Gated, it is claimable for the whole walk. This is precisely the coverage `asan.yml`'s
owner-gate row exists to provide.

**Comparison with the other #366 walks.**

| walk | protection | added by |
|---|---|---|
| `mi_heap_dump_json` | `MI_GATE_ENTER(self)` (`heap-dump.c:238`) + `MI_PARK_SWEEPING` claims on foreign theaps | `d21ee2ff` |
| `mi_heap_visit_blocks` / `mi_theap_visit_blocks` / memory-events walk | `MI_GATE_ENTER(self)` (`arena.c:3054`, `theap.c`, `memory-events.c:398`) | `90941fc3` |
| `mi_on_thread_idle` | `MI_GATE_ENTER` (`scavenger.c:810`) | `90941fc3` |
| **`_mi_purge_holes_report_collect`** | **`theaps_lock` only** (`page-holes.c:1239`) | — |

**Comment to fix either way.** `page-holes.c:905-906` and `:1218-1220` say the reporter
"never touches a free list"; `mi_holes_mark_free_list` (`page-holes.c:927-935`) reads all
three lists in place. Reword to "reads, never mutates".

**Ruled out.** A non-racy explanation: the stack is `_mi_theap_visit_pages` over the thread's
own theaps, not the abandoned pass (`_mi_arenas_holes_report`, which claims each page before
reading it, `arena.c:1508-1523`); and discarded blocks are unlinked from `page->free`
*before* the discard (`page-holes.c:633-648`) and relinked only after `_mi_os_reuse`
(`page-holes.c:423`), so a discarded block is never on a list the reporter walks.

## B. `test-heap-dump-json` — `_mi_ptr_page` on a block of a page that was just freed (guarded)

`run-linux`, run
[34055444934 attempt 1](https://github.com/zackees/mimalloc-pprof/actions/runs/34055444934),
`guarded::test-heap-dump-json [sample-rate-1]`, exit code -6:

```
mimalloc: assertion failed: at "include/mimalloc/internal.h":1113, _mi_ptr_page
  assertion: "p==NULL || mi_is_in_heap_region(p)"
```

The `guarded` bundle is `Debug + MI_PPROF=ON + MI_GUARDED=ON`, **ungated**.

**Where `_mi_ptr_page` is reached.** `mi_heap_dump_json` → `mi_heap_visit_blocks` →
`_mi_heap_visit_blocks` iterates the heap's `pages` bitmap with `claim_pages=false`
(`src/arena.c:2960-2969` — the unclaimed branch takes `mi_arena_page_at_slice` directly,
while the claiming branch has an explicit "gone: freed by a concurrent `mi_free`" NULL
check) → `_mi_theap_area_visit_blocks` walks `page->free` and `page->local_free` with
`mi_block_next` (`src/theap.c:760-762`) → `mi_block_next`'s corruption check calls
`mi_is_in_same_page(block, next)` (`internal.h:1794`) → `_mi_ptr_page(block)`
(`internal.h:1743`) → the assertion. `mi_is_in_heap_region(p)` is a page-map lookup,
`_mi_safe_ptr_page(p) != NULL` (`src/page-map.c`).

**What makes a good block address fail that lookup.** The owner (`main`, freeing
`heap1_blocks`) frees a page's last block; `_mi_page_retire` keeps the page only when it is
the only page of its queue and otherwise calls `_mi_page_free(page, pq)` immediately
(`src/page.c:512-532`); that reaches `_mi_page_map_unregister(page)` (`src/arena.c:1142`).
The dumper read the `pages` bit before the free, then reads `page->free` — still a
valid-looking block address inside the slice — and `_mi_ptr_page(block)` now finds a NULL
page-map entry. The page memory is still mapped, so this is a debug assertion, not a fault;
a release build reads stale data and moves on.

**Why guarded, and why only sample-rate-1.** The dump reports block bases `pstart + i*bsize`,
never the offset user pointer, so the guarded *offset* is irrelevant — my first guess there
was wrong. What guarded changes is page turnover: every allocation becomes
`_mi_align_up(bsize + os_page_size, os_page_size)` (`src/alloc.c:1035`), so heap1's small
blocks become multi-KiB blocks spread over many pages of a few blocks each, and the main
loop's frees empty several of them — each one an immediate `_mi_page_free` racing the
concurrent dumps. Without guarding, those blocks share a page the loop never empties, so no
page is freed and the walk cannot reach an unregistered page.

**The test's premise does not hold in this variant.** `test/test-heap-dump-json.c:253-254`
says the concurrent phase keeps pages non-empty "(only half of heap1's blocks are freed) so
this does not exercise the documented page-retirement race in #78, just the common 'read
while write' case". With `MIMALLOC_GUARDED_SAMPLE_RATE=1` it exercises exactly that race.
(`#78` here is the repo issue those thread-safety notes are hung off — see
`src/heap-dump.c:19` and `:147`, which record that `mi_heap_visit_blocks` is documented as
not thread-safe.)

**The claim protocol is vacuous in this test.** `mi_dump_claim_theaps`
(`src/heap-dump.c:174-201`) claims only PARKED tlds and walks RUNNING ones unclaimed
(`:195`); in an ungated build a thread that never called `mi_on_thread_idle_start` is always
RUNNING, so nothing is claimed here and the dump is the plain upstream walk.
`MI_DUMP_MAX_CLAIMS=64` is not a factor at two threads — my first reading blamed the cap, and
that was wrong.

## Frequency and reproduction

- PR #372, attempt 1 of both workflows: **both green**. Attempt 2 (identical C — the push in
  between touched only `ci/tests/` and `docs/`): **both failed**. Rerun of attempt 2's failed
  jobs: **both green**. `main`'s last eight `c-unit.yml` and `asan.yml` runs: green.
- **Not reproduced locally** (x86_64, 16 cores, clang 21.1.8, `main` at `4ef418b8`):
  - A: 200 runs of `mimalloc-test-purge-holes` built `-DCMAKE_BUILD_TYPE=Debug -DMI_PPROF=ON
    -DMI_DEBUG_FULL=ON -DMI_OWNER_GATE=ON -DMI_TRACK_ASAN=ON` (8 concurrent × 25), plus 160
    more pinned to 4 CPUs with `taskset -c 0-3` to imitate the runner — all passed.
  - B: 240 runs of `mimalloc-test-heap-dump-json` built `-DCMAKE_BUILD_TYPE=Debug
    -DMI_PPROF=ON -DMI_GUARDED=ON` with `MIMALLOC_GUARDED_SAMPLE_RATE=1` (6 concurrent × 40)
    — all passed.

Both windows are tiny (A: between an unpoison and the next instruction; B: between a bitmap
read and a free-list read, per emptied page), and A additionally needs the scavenger to wake
inside the report. Deterministic reproducers that do not depend on scavenger timing:

- **A:** gated ASan build; main thread loops `_mi_purge_holes_report_collect` over a heap laid
  out as in `test_report_is_read_only`, while a second thread loops `mi_purge_all(true)`,
  which takes the same claim and runs the same `_mi_thread_idle_work_ex` foreign door
  (`src/purge-all.c:165-209`). The reporting thread is PARKED for every report, so every
  `mi_purge_all` claims it.
- **B:** guarded Debug build, `MIMALLOC_GUARDED_SAMPLE_RATE=1`; dump thread loops
  `mi_heap_dump_json(true, ...)` while the owner loops alloc-then-free of a single block on a
  heap that already holds a second page in the same bin, so every free is an immediate
  `_mi_page_free`.

## Why it matters even though it is rare

Under ASan / `MI_DEBUG` both are aborts. In a release build A silently truncates the
reporter's free-list walk (the holes report under-counts and nobody notices) and B reads a
freed page's stale `free` list. The #366 gate/claim protocol exists to make these walks safe
and one walk was left out of it; B is a pre-existing documented race that a debug assertion
now turns into a red gate.

## Suggested work (ordered)

1. **A, fix:** gate the reporter. In `_mi_purge_holes_report_collect`
   (`src/page-holes.c:1227-1257`) take `MI_GATE_ENTER(theap0)` before the `theaps_lock` walk
   and `MI_GATE_LEAVE(tld)` after, exactly as `mi_on_thread_idle` does
   (`src/scavenger.c:803-813`). Being RUNNING is sufficient: a claimed sweep needs PARKED, and
   the abandoned pass already claims pages (`arena.c:1508-1523`). Add the site to the §5.1
   list in `docs/purge-all-implementation.md`.
2. **A, comment:** `src/page-holes.c:905-906` and `:1218-1220` — "never touches a free list"
   → "reads the three free lists in place and never mutates them".
3. **B, decide the contract, then one of:**
   - (a) keep the dump best-effort but never dereference an unpinned page: in the unclaimed
     branch (`src/arena.c:2960-2969`) pin each page for its visit with the existing bit-clear
     protocol from `mi_heap_visit_page_claim` (clear the `pages` bit → visit → set it), since
     the owner's `mi_bitmap_clear_once_set` in `mi_arenas_page_free_prim` waits while the bit
     is held, so `_mi_page_map_unregister` cannot run under the walk. Apply the same to
     `mi_heap_visit_os_pages`' no-claim path (`arena.c:2977-2992`). Residual: a torn `next`
     still hits `mi_block_next`'s "corrupted free list" path (`internal.h:1795-1797`), which
     is a message, not an abort — acceptable for a documented best-effort walk. Or
   - (b) keep the walk as is and fix the test: under `MIMALLOC_GUARDED_SAMPLE_RATE=1` the
     concurrent phase of `test/test-heap-dump-json.c:253+` violates its own stated premise, so
     skip that phase when `mi_option_get(mi_option_guarded_sample_rate) == 1`, or free only
     blocks whose page cannot empty. That removes the red gate but leaves the assertion
     reachable by any user dumping a heap another thread frees into.
4. **Audit:** the remaining unclaimed page-walks of this shape are the two in 3(a); every
   other `_mi_theap_visit_pages` / `mi_heap_visit_blocks` / `_mi_theap_area_visit_blocks`
   caller is already gated (table above). Record the result in
   `docs/purge-all-implementation.md` so the next walk added is checked against it.
5. **Stress tests** in the `RUN_SERIAL` group of `CMakeLists.txt`: the two reproducers above
   (A gated-only, B guarded-only), so this stops depending on CI luck. A's test should assert
   that the sweep actually ran, so it cannot pass vacuously.

## Related

- #366 (owner gate) is what A is a gap in; `d21ee2ff` and `90941fc3` are the commits that
  gated every other walk.
- The thread-safety of `mi_heap_visit_blocks` is documented as best-effort in
  `src/heap-dump.c:19,147` and `include/mimalloc.h`; B is that documented race reaching an
  assertion.
- #371 (DHAT hot-path cost) is unrelated as a cause, but its comment thread carries the same
  observations.
- Separately: the win-gnu `test-purge-all` failure that is red on `main` is **also flaky**, not
  deterministic — it passed twice and failed once across #372's runs. Whoever fixes that should
  expect an intermittent reproducer.

zackees:
Closes #374.

## Approved resolution

Implements the ownership-protected, explicitly incomplete diagnostic capture approved in [the issue discussion](https://github.com/zackees/mimalloc-pprof/issues/374#issuecomment-5650195758), with the [performance investigation and recommendation](https://github.com/zackees/mimalloc-pprof/issues/374#issuecomment-5650207065).

- Gate the holes reporter against foreign collection; correct the free-list-read comments and gate inventory.
- Replace heap-dump use of the general best-effort visitor with a dedicated ownership-protected capture. Never read an unclaimed RUNNING/SWEEPING owner's mutable fields. Cover arena pages, live OS pages, abandoned OS pages, and metadata ownership without changing retirement ordering.
- Report `complete`, `skipped_pages`, and `busy_theaps`. Complete means no observed omissions, not an atomic process-wide instant. Ungated foreign owners require successful cooperative parking for coverage; merely sleeping is insufficient. No new public complete-snapshot API and no mandatory owner gate.
- Capture page/block records and serialize using amortized raw-OS scratch. Allocate the `mi_free`-compatible result only after all capture claims and registry locks are released. Return NULL on allocation failure, never truncated JSON.
- Preserve exact visitor assertions, guarded concurrency tests, public generic visitor preconditions, and default allocation/free fast paths. Refresh Rust vendor and document its wrapper contract in separate Rust-only commits.

## Regression coverage

- Deterministic reporter-vs-sweeper exclusion, followed by an actual successful foreign purge (RED before gate fix, GREEN after).
- Busy-owner omissions vs exact parked-owner counts (old dump RED on explicit coverage assertion).
- Arena and forced-OS variants; concurrent allocation/free page-retirement churn.
- 40,000-block buffer growth with no hooked resize; 70 parked owners (no fixed 64-owner cap).
- 70 abandoned singleton pages, capture allocation failures including a partially claimed OS batch, formatting failure, successful retry, and cross-thread frees/cleanup.
- Existing JSON test retains guarded concurrent frees and now verifies block/page usable-size consistency.

The historical Rust assertion was not reproduced at its exact failing revision. That revision does contain a with-blocks heap-dump test, contrary to the original comment. This change removes both concrete hazards (unowned foreign reads and same-heap buffer growth during capture); the plain no-default-features Rust tests pass. No weakened assertion or skipped guarded test is used as a substitute for safety.

## Local evidence

- Final source `ef905c90`: gated Debug diagnostic/holes tests 5/5; ungated PPROF=OFF Debug diagnostic/JSON tests 3/3, including guarded sample rate 1; real Clang ASan gated tests 5/5; C++-compiled focused diagnostic/JSON tests 3/3.
- Rust plain lib tests 8/8 and default lib tests 9/9; `xtask check`, Rust binding surface check, internal-state inventory and positive controls, and `git diff --check` pass. `48467740` adds only the audited scratch/result allocation inventory and documentation.
- `ci/check_fastpath_identity.py --base 7d8e5e86` passes at final source: default `mi_malloc`, `mi_zalloc`, `mi_free`, `mi_heap_malloc_small`, `mi_malloc_small` are byte-identical after documented relocation normalization; gated positive control differs in all five. This is evidence for those five entry points, not a whole-program speed claim.
- Local GCC Release quiescent 40,000 x 32-byte target-heap benchmark, one warmup + 20 dumps, scavenger disabled: old median 5.361 ms; new 6.380 ms (~19% diagnostic-call cost). Both capture exactly 40,000/40,000 blocks. JSON sizes 763,503 vs 762,894 bytes. Timing is machine/workload-specific; no unreliable process RSS claim is made.
- Broader optional GCC Debug `MI_USE_CXX=ON` build encounters an existing untouched `test-purge-holes.c` goto-crossing-initialization error; focused C++ diagnostics pass. Required native MSVC and cross-platform gates must still pass before merge.

Independent pre-push review found and verified fixes for fresh-thread OOM handling, foreign gate-depth access ordering, OS-batch/error coverage, and the guarded block-address fixture. All findings resolved.

review agents launched: 1
clud-review: clean

## Merge gates

Awaiting this PR's Linux ON/OFF, ASan, native **cl** MSVC build/run, win-gnu and clang-cl Windows bundles, Rust, and both macOS cross-build architectures/selective decision checks. Do not merge by bypassing a required gate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1215,6 +1215,27 @@ if (MI_BUILD_TESTS)
     set_tests_properties(test-heap-teardown PROPERTIES TIMEOUT 300)
   endif()
 
+  # #374 uses native threads on Windows and pthreads elsewhere. Keep these outside
+  # the POSIX-only tests: both MSVC and win-gnu must execute the ownership regressions.
+  # Internal page state requires the static library and its exact compilation defines.
+  if(MI_BUILD_STATIC AND NOT MI_DEBUG_TSAN)
+    add_executable(mimalloc-test-diagnostic-walks test/test-diagnostic-walks.c)
+    target_compile_definitions(mimalloc-test-diagnostic-walks PRIVATE ${mi_defines} MI_STATIC_LIB)
+    target_compile_options(mimalloc-test-diagnostic-walks PRIVATE ${mi_cflags})
+    target_include_directories(mimalloc-test-diagnostic-walks PRIVATE include)
+    target_link_libraries(mimalloc-test-diagnostic-walks PRIVATE mimalloc-static ${mi_libraries})
+    if(NOT WIN32)
+      find_package(Threads REQUIRED)
+      target_link_libraries(mimalloc-test-diagnostic-walks PRIVATE Threads::Threads)
+    endif()
+    add_test(NAME test-diagnostic-walks COMMAND mimalloc-test-diagnostic-walks)
+    set_tests_properties(test-diagnostic-walks PROPERTIES TIMEOUT 120
+                         ENVIRONMENT "MIMALLOC_SCAVENGER=0")
+    add_test(NAME test-diagnostic-walks-os COMMAND mimalloc-test-diagnostic-walks)
+    set_tests_properties(test-diagnostic-walks-os PROPERTIES TIMEOUT 120
+                         ENVIRONMENT "MIMALLOC_SCAVENGER=0;MIMALLOC_DISALLOW_ARENA_ALLOC=1")
+  endif()
+
   # pthread_atfork fork-safety handlers (issue #270, Bun parity P5). POSIX-only --
   # both sources compile to a Windows-skip stub via #ifdef _WIN32, but there is no
   # pthread/fork() to link against on MSVC/MinGW, so only build+register them off
@@ -1335,17 +1356,6 @@ if (MI_BUILD_TESTS)
     # (`MI_BUILD_STATIC=OFF`) skips it rather than failing to link.
     if(MI_BUILD_STATIC AND NOT MI_DEBUG_TSAN)
       add_executable(mimalloc-test-purge-holes test/test-purge-holes.c)
-      add_executable(mimalloc-test-diagnostic-walks test/test-diagnostic-walks.c)
-      target_compile_definitions(mimalloc-test-diagnostic-walks PRIVATE ${mi_defines} MI_STATIC_LIB)
-      target_compile_options(mimalloc-test-diagnostic-walks PRIVATE ${mi_cflags})
-      target_include_directories(mimalloc-test-diagnostic-walks PRIVATE include)
-      target_link_libraries(mimalloc-test-diagnostic-walks PRIVATE mimalloc-static ${mi_libraries} Threads::Threads)
-      add_test(NAME test-diagnostic-walks COMMAND mimalloc-test-diagnostic-walks)
-      set_tests_properties(test-diagnostic-walks PROPERTIES TIMEOUT 120
-                           ENVIRONMENT "MIMALLOC_SCAVENGER=0")
-      add_test(NAME test-diagnostic-walks-os COMMAND mimalloc-test-diagnostic-walks)
-      set_tests_properties(test-diagnostic-walks-os PROPERTIES TIMEOUT 120
-                           ENVIRONMENT "MIMALLOC_SCAVENGER=0;MIMALLOC_DISALLOW_ARENA_ALLOC=1")
       target_compile_definitions(mimalloc-test-purge-holes PRIVATE ${mi_defines})
       target_compile_options(mimalloc-test-purge-holes PRIVATE ${mi_cflags})
       target_include_directories(mimalloc-test-purge-holes PRIVATE include)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -754,6 +754,7 @@ endif()
 # set language for source files now
 if(MI_USE_CXX)
   set_source_files_properties(${mi_sources} PROPERTIES LANGUAGE CXX )
+  set_source_files_properties(test/test-diagnostic-walks.c PROPERTIES LANGUAGE CXX)
   # #272 (P7b): test-purge-holes.c and test-profile-race.c reach `mimalloc/internal.h`
   # (`_mi_ptr_page`, `mi_page_block_is_purged`, `_mi_os_page_size`, ...), which is NOT wrapped
   # in `extern "C"`. With MI_USE_CXX the library's symbols are C++-mangled, so a C test cannot
@@ -1334,6 +1335,17 @@ if (MI_BUILD_TESTS)
     # (`MI_BUILD_STATIC=OFF`) skips it rather than failing to link.
     if(MI_BUILD_STATIC AND NOT MI_DEBUG_TSAN)
       add_executable(mimalloc-test-purge-holes test/test-purge-holes.c)
+      add_executable(mimalloc-test-diagnostic-walks test/test-diagnostic-walks.c)
+      target_compile_definitions(mimalloc-test-diagnostic-walks PRIVATE ${mi_defines} MI_STATIC_LIB)
+      target_compile_options(mimalloc-test-diagnostic-walks PRIVATE ${mi_cflags})
+      target_include_directories(mimalloc-test-diagnostic-walks PRIVATE include)
+      target_link_libraries(mimalloc-test-diagnostic-walks PRIVATE mimalloc-static ${mi_libraries} Threads::Threads)
+      add_test(NAME test-diagnostic-walks COMMAND mimalloc-test-diagnostic-walks)
+      set_tests_properties(test-diagnostic-walks PROPERTIES TIMEOUT 120
+                           ENVIRONMENT "MIMALLOC_SCAVENGER=0")
+      add_test(NAME test-diagnostic-walks-os COMMAND mimalloc-test-diagnostic-walks)
+      set_tests_properties(test-diagnostic-walks-os PROPERTIES TIMEOUT 120
+                           ENVIRONMENT "MIMALLOC_SCAVENGER=0;MIMALLOC_DISALLOW_ARENA_ALLOC=1")
       target_compile_definitions(mimalloc-test-purge-holes PRIVATE ${mi_defines})
       target_compile_options(mimalloc-test-purge-holes PRIVATE ${mi_cflags})
       target_include_directories(mimalloc-test-purge-holes PRIVATE include)
@@ -2054,6 +2066,8 @@ endif()
 # (MI_PPROF, non-Windows, ...), and set_tests_properties() on a test that was never
 # registered is a hard CMake error.
 set(mi_serial_tests
+  test-diagnostic-walks
+  test-diagnostic-walks-os
   test-osx-zone-introspect-remote
   test-memory-gate
   test-degenerate
--- a/ci/internal-state-inventory.json
+++ b/ci/internal-state-inventory.json
@@ -329,22 +329,42 @@
       ]
     },
     {
+      "id": "heap-dump-capture-scratch",
+      "path": "src/heap-dump.c",
+      "callee": "_mi_os_alloc",
+      "args": "ctx->subproc,sizeof(*chunk),&memid",
+      "occurrence": 1,
+      "owner": "operation-local mi_dump_ctx_t",
+      "initial_owner": "raw OS allocation; never a hooked allocator heap",
+      "owner_transitions": "initialized chunk is prepended to ctx->chunks; records and text borrow bump-allocated extents",
+      "destroyable_by": "mi_dump_dispose on every success or failure path",
+      "lifetime": "one mi_heap_dump_json call, including capture and serialization",
+      "logical_size": "sizeof(mi_dump_chunk_t), including 64 KiB of scratch data",
+      "usable_size": "only sizeof(*chunk) is used and passed with its memid to _mi_os_free",
+      "cleanup": "capture failure first releases page ownership and locks; mi_dump_dispose frees every chunk even when formatting or final result allocation fails",
+      "publication": "chunk memid, used and next initialized before private ctx->chunks update; never published to other threads",
+      "locks": [
+        "capture: subproc heaps_lock and heap theaps_lock; may also hold a page pin, owner claim, metadata lock, or OS abandoned-pages lock",
+        "serialization: no traversal lock or owner claim; raw OS allocation cannot reenter hooked allocation"
+      ]
+    },
+    {
       "id": "heap-dump-json-buffer",
       "path": "src/heap-dump.c",
-      "callee": "mi_rezalloc",
-      "args": "hbuf->buf,newsize",
+      "callee": "mi_malloc",
+      "args": "ctx.text_size+1",
       "occurrence": 1,
       "owner": "mi_heap_dump_json caller",
       "initial_owner": "calling thread's default heap",
-      "owner_transitions": "returned to the dump caller; moving growth preserves old buffer on failure, matching stats.c's mi_json_buf_expand this mirrors",
+      "owner_transitions": "allocated once after capture and serialization; copied from raw-OS scratch then returned to the dump caller",
       "destroyable_by": "caller via mi_free (Bun's BunJSCModule.h frees the same way after JSONParse)",
       "lifetime": "until caller invokes mi_free",
-      "logical_size": "initial good size near 12 KiB, then doubled",
-      "usable_size": "allocator size class; hbuf->size retains requested logical capacity",
-      "cleanup": "caller owns returned buffer and frees with mi_free; a growth failure mid-print leaves the buffer at its last successfully-written prefix (same discipline as src/stats.c's mi_json_buf_expand)",
-      "publication": "hbuf pointer and size update only after successful rezalloc",
+      "logical_size": "ctx.text_size + 1, with overflow checked while accumulating text",
+      "usable_size": "allocator size class; copy only ctx.text_size bytes and one NUL terminator",
+      "cleanup": "caller frees successful result with mi_free; any allocation failure returns NULL after freeing raw-OS scratch, never a truncated JSON prefix",
+      "publication": "returned only after every text chunk is copied and the NUL terminator is written",
       "locks": [
-        "none; buffer is operation-local"
+        "all diagnostic traversal locks and owner claims released before mi_malloc; ordinary allocation may acquire its usual allocator locks"
       ]
     },
     {
--- a/docs/internal-state-diagnostics.md
+++ b/docs/internal-state-diagnostics.md
@@ -47,6 +47,8 @@ fields.
 | `profiler-proto-modules` | raw OS -> dump-local modules | one protobuf dump | fixed module array / same cleanup expression | never global; every exit frees |
 | `profiler-proto-pc-table` | raw OS -> dump-local hash table | one protobuf dump | power-of-two table / same cleanup expression | zero before use; every exit frees |
 | `stats-json-buffer` | default heap -> stats caller | caller lifetime | requested capacity / allocator class | pointer/size update after successful growth; caller frees |
+| `heap-dump-capture-scratch` | raw OS -> private dump capture/serialization arena | one JSON dump; dispose on every exit | header + 64 KiB / exact retained size | initialize then link; capture may hold ownership/registry locks; no hooked allocation |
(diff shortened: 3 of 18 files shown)


zackees:
CI follow-up: the first head passed all **307 glibc + 72 musl parallel runtime tests**, all four ASan configurations, native cl MSVC build/run (including owner-gate), Rust, and macOS cross-build gates. `run-linux` then failed its release-symbol firewall because the new production helper `_mi_heap_visit_diagnostic` matched the existing `_mi_.*diagnostic` pattern reserved for MI_DEBUG-only hooks.

Fixed in `ec0d30a8` + separate Rust regeneration `9d43f133`: rename the private production entry point to `_mi_heap_visit_capture`. No safety/allocator logic or CI-checker change. Local nm reproduces the old match and confirms no forbidden symbols in both PPROF=ON and OFF Release archives after the rename. Release-equivalence positive controls, focused gated tests, vendor check and independent narrow review pass. Final-head fastpath identity/positive control also pass again against `7d8e5e86`.

Pushing the corrected head for fresh required gates; no failed test is being retried into green and no gate is being bypassed. Full local suites from the previous source revision also passed: gated C 65/65, PPROF=OFF C 59/59, full Rust workspace and allocator-only crate suites.

zackees:
Coverage audit caught and fixed a registration mistake before merge: the new diagnostic test target had been placed under CMake's existing outer `if(NOT WIN32)`, despite the source having a native Windows thread implementation. The older Windows gates therefore tested the changed library/existing JSON test, but did **not** run the new diagnostic ownership fixtures.

`85c7f6ec` moves just this registration outside the POSIX block. It retains static-only/non-TSan constraints, exact library defines and MSVC C++ compilation; pthread linkage is conditional on non-Windows. Both tests stay RUN_SERIAL.

Local evidence: both **MinGW-w64 GCC 15.3 C** and **clang-cl 21.1.5 C++** cross-build the new executable successfully with MI_DEBUG_FULL=ON and MI_OWNER_GATE=ON. `ctest -N` lists `test-diagnostic-walks` and `test-diagnostic-walks-os` on both Windows targets. Linux focused tests 2/2, vendor drift check and independent follow-up review pass. This is compile/registration evidence, not a claim of Windows execution; fresh native cl and Windows bundle results are still required before merge.

No allocator or Rust source changed in this follow-up.

zackees:
Implemented the bounded-wait recommendation in #398 (`92c42f1d`).

The safe design is a **whole-capture retry from a clean boundary**, not a spin inside the per-thread walker:

- A top-level owner-gated dump attempts the capture, releases its caller gate plus every subprocess/heap lock, page pin, owner claim, and raw scratch allocation, then retries incomplete captures until one dump-wide acquisition deadline.
- The legacy `mi_heap_dump_json()` uses 100 ms. New `mi_heap_dump_json_ex(..., wait_ms)` and Rust `heap_dump_json_ex(..., wait_ms)` expose the bound; `wait_ms == 0` is one attempt.
- `MI_OWNER_GATE=OFF` remains one-shot because waiting cannot make an ordinary RUNNING foreign owner claimable.
- Calls nested inside an already-gated allocator operation are also one-shot: the dump cannot release its caller's outer gate safely.
- Timeout accounting uses overflow-free elapsed-time comparison, including `SIZE_MAX`.

Why not wait inside `mi_diag_try_tld`: capture reaches a TLD while holding traversal locks and may have arena pins. A RUNNING owner can need those same resources to finish allocation, retirement, heap creation/deletion, or teardown. Spinning there creates a lock inversion. Discarding the incomplete attempt before waiting avoids that cycle and also avoids needing a new TLD lifetime-reference protocol.

Thread initialization is safe under the existing ordering: a TLD is registered RUNNING before a theap becomes reachable under its heap lock. A thread created concurrently with the dump does not define a process-wide atomic cutoff; `complete`, `skipped_pages`, and `busy_theaps` describe the final observed attempt.

Profiler scratch and serialization buffers remain raw-OS-arena backed. Only the final caller-owned JSON string is allocated through mimalloc, after capture locks/claims are gone.

Performance: the default `MI_OWNER_GATE=OFF` allocator hot paths remain byte-identical to the pre-change base and contain no lock-prefixed instruction. Waiting runs only when a dump is requested. Gated allocation retains the already-intended per-thread owner gate; a dump may delay it only during the rare ownership handoff/capture window.

Validation includes deterministic busy-owner release, attachment to a previously unused heap between attempts (proving traversal locks were released), nested one-shot behavior, `SIZE_MAX`, retirement churn, ASan, gated/ungated Linux, Rust feature-on/off, Rust/C and Bun API surfaces, plus MinGW and clang-cl cross-builds. Independent review is clean.

Issue #415, opened 2026-09-20
perf: make the observer hooks free when disarmed (fold into the existing free branch)


zackees:
Companion to #414. #414 makes the observers opt-in because memory-events costs real instructions today. This issue removes that cost, after which memory-events could go back to default-on at **zero** fast-path cost.

Measured on `origin/main` @ 4aee27a1 (gcc 15.2, Release, `perf stat`, 20M/40M ops, +/-0.0002 instr/op; upstream `6def7be9` = floor):

| Variant | instr/op | vs upstream | memevt-attributable |
|---|---:|---:|---:|
| upstream `6def7be9` | 27.37 | - | - |
| main today | 35.30 | +7.94 | +4.52/op (9.04/pair) |
| + `mi_decl_hidden` one-liner | 34.30 | +6.94 | +3.52/op |
| V1: cold + tail-call | 33.30 | +5.94 | unchanged (removes the *pprof* frame) |
| V2: fold into the existing free branch | 30.82 | +3.46 | **~0** |
| V1+V2 | 28.82 | **+1.46** | **0** |

Static counts agree: upstream `mi_free`+`mi_malloc` = 36 instructions per pair; main = 51; V1+V2 = 38. With V2, `mi_malloc` is **byte-identical to upstream** (the disassembly differs only in three branch target addresses).

### 1. `mi_decl_hidden` on `_mi_observers_armed` (one word, do first)
`include/mimalloc/internal.h:636` declares it `extern` without `mi_decl_hidden`, so under `-fPIC` GCC routes the relaxed load through the GOT (`lea sym(%rip),%r; mov (%r),%r`) instead of `mov sym(%rip),%r`. **-1 instruction per hook site = -1.0 instr/op, 11% of the memevt cost, zero risk.** Helps the DHAT hooks too.

### 2. Cold-outline the slow paths (~1 day)
Mark `_mi_memevt_on_*_slow` `noinline`+`cold` and restructure the call sites so the compiler emits a sibling call. `mi_free` then keeps no stack frame (today it does `push %rbp / mov / sub $0x10 / ... / leave`, and both out-of-line calls spill through it). Worth **2 instr/pair**, and the saving is really a **pprof** fix — it is the adjacent `has_metadata` hook that forces the frame — so it lands independently of anything memevt.

### 3. Fold the observer check into the branch `mi_free` already takes (~2-3 days)
[...]

Issue #337, opened 2026-09-04
feat: restore zero-tracking (zalloc skips memset after a zeroing purge) — regressed out of the tree by the #80 pin bump


zackees:
Filed so the README feature-comparison table's `Zero-tracking` cell has a home. **Bun has it; this fork does not.**

## State
`include/mimalloc.h:609` is explicit:

```
mi_option_purge_zeroes,  // DEAD since #80: the implementation added by #79 went away with the v3 pin bump and was never
                         // restored (see issue #67). The option slot is kept -- never renumber -- and setting
                         // MIMALLOC_PURGE_ZEROES still parses, it just has no effect.
```

`_mi_os_purge_zero` does not exist anywhere in `src/` or `include/` here.

## What Bun has at `b20b60d9`
- `include/mimalloc/internal.h:229` — `bool _mi_os_purge_zero(mi_subproc_t*, void* p, size_t size, size_t stat_size, bool* is_zero);`
- `src/os.c:658` — the implementation
- `src/arena.c:2375` — the call site that records `is_zero` so a later `mi_zalloc` can skip its `memset`

Note Bun carries the *mechanism* without an option knob; the `purge_zeroes` option slot is this fork's.

## Why this is a fresh issue rather than reopening #67
[#67](https://github.com/zackees/mimalloc-pprof/issues/67) was closed **COMPLETED** — the work landed in #79. The regression happened afterwards, in the #80 overlay pin bump, so #67's closure is accurate history and this is a different event: not "never implemented", but "implemented, then lost in a rebase and not noticed".

## What "done" looks like
- `_mi_os_purge_zero`-equivalent plumbing restored, `mi_option_purge_zeroes` live again (same slot — never renumber).
- #67's rule still binds: **any flag that ships must have a CI job running the full suite with it ON.** `.github/workflows/zero-tracking.yml` and `ctest -R zero-tracking` already exist for exactly this and are currently exercising a no-op.
- A regression guard so the next overlay bump cannot silently drop it again — the failure mode here was that the correctness tests still pass when the feature does nothing.
- Flip the README feature table's `Zero-tracking` row for `mimalloc-pprof` from ❌ to ✅ (`docs/allocator-features.json`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YT4jVokb2gdT8ngFrH8i8S

PR #416, opened 2026-09-20
feat!: every observability subsystem opt-in; the default build is a plain fast allocator (0.12.0)

### LINKED ISSUE #414: feat: every observability subsystem opt-in — default build is a plain fast allocator (0.12.0)
Owner decision 2026-09-19: **every observability subsystem is opt-in. The default build is a plain fast allocator.** This follows #411 (DHAT) and applies the same shape to the rest, on both surfaces (Cargo features *and* CMake options), and ships as **0.12.0**.

## Why (measured, not estimated)

Release builds, gcc 15.2, identical flags; upstream = pinned `6def7be9` (218 KB stripped, ~50–55 instructions per malloc/free pair). Instruction counts from `perf stat` over a 20M/40M-op malloc/free benchmark, reproducible to ±0.0002 instr/op.

| Subsystem | | Compiled in, disabled | Compiled out |
|---|---|---:|---:|
| pprof (stopped) | Memory | 28.9 KB + 16 B/page | 0.9 KB stubs |
| | CPU | +3.4 instr/pair (+10% cycles) | 0 |
| memory-events | Memory | 4.0 KB | ~0.3 KB stubs |
| | CPU | **+9 to +13 instr/pair (18–26%)** | 0 |
| DHAT (stopped) | Memory | 9.8 KB | 0.5 KB stubs |
| | CPU | 0 (byte-identical) | 0 |
| heap snapshot | Memory | 21.9 KB | ~0.1 KB stubs |
| | CPU | 0 | 0 |
| heap dump | Memory | 3.8 KB | ~0.1 KB stubs |
| | CPU | 0 | 0 |
| owner-gate | Memory | 16.4 KB | 0 (already off) |
| | CPU | +95 instr/pair (+206% cycles) | 0 |

Allocator behavior stays compiled in unconditionally: scavenger, page-holes, purge-all, zero-tracking, fork handlers (all 0 instr/pair on malloc/free; Bun parity).

### 1. Cargo features (`rust/mimalloc-pprof/Cargo.toml`)

```toml
default = []
pprof = []                    # MI_PPROF=1
memory-events = []            # MI_MEMEVT=1   (new)
diagnostics = []              # MI_DIAGNOSTICS=1 (new): heap snapshot + heap dump
dhat = ["memory-events"]      # MI_DHAT=1; DHAT dispatches through the memevt slow paths
owner-gate = []               # MI_OWNER_GATE=1
full = ["pprof", "memory-events", "diagnostics", "dhat", "owner-gate"]
```

`build.rs` maps each to its define (0 or 1, always defined), exactly as `dhat` does today.

### 2. CMake options (`CMakeLists.txt`)

- `MI_PPROF` ON -> **OFF**
- new `MI_MEMEVT` **OFF**, new `MI_DIAGNOSTICS` **OFF**
- `MI_DHAT`, `MI_OWNER_GATE` already OFF
- `include/mimalloc/types.h` fallback defines must match (`#define MI_PPROF 0`, `MI_MEMEVT 0`, `MI_DIAGNOSTICS 0`), so a direct `src/static.c` compile also gets the minimal build.

### 3. Compile-out work

- **memory-events:** guard the inline hooks in `include/mimalloc/internal.h` with `#if MI_MEMEVT || MI_DHAT` (DHAT dispatches from memevt's slow paths at `src/memory-events.c:284-295,328-337,349-359,371-380`, so the hook sites must survive whenever `MI_DHAT=1`). Add an `#else` stub block for the public exports. `MI_OBSERVERS_MEMEVT_UNRESOLVED` must drop out of `MI_OBSERVERS_INITIAL` the way DHAT's already does (`internal.h:628-633`), or the fast path takes the slow branch forever. **Keep `mi_unwrapped_malloc/_free/_realloc` real** — they are raw-OS wrappers, unrelated to tracking.
- **diagnostics:** stub `mi_heap_snapshot`, `mi_heap_snapshot_to_file`, `_mi_heap_snapshot_on_exit`, `mi_heap_dump_json{,_ex}`, `mi_heap_get_seq`. Keep `mi_option_snapshot_on_exit` present in every build (Bun parity contract).
- **pprof:** already gated; additionally put `#if MI_PPROF` on `mi_page_t::metadata` / `has_metadata` (`types.h:465-466`) so a profiler-off page header drops 192 -> 176 B. Check `test/test-bun-surface.cpp` and `rust/.../tests/t19_layout.rs` first.

### 4. API surface must not change

Every `mi_*` export and every Rust wrapper stays present in every configuration, as stubs returning false/None — the template is `src/profile.c:1126-1150`. This keeps `ci/check_rust_surface.py`, `layout_probe.c`, `t19_layout.rs` and the README API table valid, and means `--no-default-features` never fails to *compile* a downstream crate. Extend `tests/feature_contract.rs` rather than inventing a second pattern. Add `required-features` to the `[[test]]` entries that need them (`t18_heap_snapshot`, `t24_purge_all` are unconditional today).

### 5. CI must not lose coverage (the #411 lesson)

Every workflow row that relied on a default must now pass the flag explicitly, or that subsystem silently stops being tested. Audit `c-unit.yml`, `windows-bundles.yml`, `macos-bundles.yml`, `asan.yml`, `cross.yml`, `rust-native.yml`, plus `ci/verify_local.py`, `ci/dev_linux.py`, `ci/build_pprof_tax_configurations.py`, `ci/check_crate_package.py`, `ci/check_fastpath_identity.py`.

Test the features individually and all-on; do not build the cross product (#307). The one pair with a proven interaction is **memory-events x dhat**, because `MI_MEMEVT=0 MI_DHAT=1` is where the `#if MI_MEMEVT || MI_DHAT` guard is load-bearing. Suggested rows: `minimal` (everything off), `release` (full), `memevt-only`, `dhat-no-memevt`, plus the existing `dhat-on`, `gated`, `debug-full`, `guarded`, `shared`, `musl`, `isa-*`, `diag-*`. Add `owner-gate` to `rust-native.yml`'s feature-combination lane, which does not cover it today.

### 6. New gate

Extend `ci/check_fastpath_identity.py` with a third comparison: the **minimal build's `mi_malloc`/`mi_free` must be byte-identical to upstream `6def7be9`**. The prototype confirms `mi_malloc` already is once the memevt hook is out. This turns "the allocator-only build has no tax" into a check instead of a claim.

### 7. Docs and version

- `rust/mimalloc-pprof/Cargo.toml` version -> **0.12.0**, plus CHANGELOG entry stating the breaking default change and how to restore old behavior (`features = ["full"]`).
- README: Quick start must show the opt-in features (gated by `ci/check_rust_surface.py`); document that Cargo features are additive and unified, so a dependency that pulls the crate with default features re-enables them (`cargo tree -e features` to check).
- Fix wrong claims found while measuring: `Cargo.toml` says owner-gate costs "a few percent" (measured +206% cycles in a tight loop); README:641-643 says memory-events costs "one relaxed flag check and nothing else" (it is 9-13 instructions per pair including a forced stack frame); README:699 says 61 enumerators / 13 fork options at 47-59 (actually 62 / 14 / 47-60, `mi_option_snapshot_on_exit` missing); README:706 calls `purge_zeroes` "dead since #80" (restored by #337).

## Acceptance criteria

- [ ] `default = []`; every subsystem opt-in on both surfaces; `full` convenience feature exists
- [ ] `cargo build -p mimalloc-pprof` with no features compiles, links, and passes its tests; so do `--features full` and each feature alone
- [ ] Public API present in every configuration; `ci/check_rust_surface.py` green with and without features
- [ ] `ctest` green for: everything off, everything on, `MI_MEMEVT=0 MI_DHAT=1`, `MI_MEMEVT=1 MI_PPROF=0`
- [ ] All CLAUDE.md merge gates green, with explicit flags wherever a default was relied on
- [ ] `check_fastpath_identity.py` proves the minimal build byte-identical to upstream
- [ ] Version 0.12.0 + CHANGELOG; docs corrected
- [ ] Commits split per CLAUDE.md rule 2 (C-core / rust/ / ci+docs)

## Not in scope

Making the hooks themselves free (fold the observer check into the branch `mi_free` already takes; cold-outline the slow paths; the `mi_decl_hidden` one-liner). That is tracked in **#415** — with it, memory-events could return to default-on at zero cost.

zackees:
## What

**Every observability subsystem is opt-in. The default build is a plain fast allocator.** Owner decision 2026-09-19. This is #411's shape (DHAT) applied to the rest, on both surfaces, shipped as **0.12.0**.

| Surface | Before | After |
|---|---|---|
| Cargo | `default = ["pprof"]` | `default = []`; `pprof`, `memory-events`, `diagnostics`, `dhat` (implies `memory-events`), `owner-gate`, `full` |
| CMake | `MI_PPROF` ON | `MI_PPROF` **OFF**, new `MI_MEMEVT` **OFF**, new `MI_DIAGNOSTICS` **OFF**; `MI_DHAT`/`MI_OWNER_GATE` already OFF |
| `types.h` fallbacks | `MI_PPROF 1` | all five `0`, so a direct `src/static.c` compile (Bun's) gets the same minimal build |

**The API does not change.** Every `mi_*` export and every Rust wrapper is present in *every* configuration as a stub that reports the subsystem off — template: the `#else` block at the end of `src/profile.c`. `ci/check_rust_surface.py`, `layout_probe.c`, `t19_layout.rs` and the README API table stay valid, and `default-features = false` never fails to *compile* a downstream crate.

## Why (measured, from #414)

Release builds, gcc 15.2, identical flags; upstream = pinned `6def7be9`. Instructions per malloc/free pair from `perf stat` over a 20M/40M-op benchmark.

| Subsystem | | Compiled in, disabled | Compiled out |
|---|---|---:|---:|
| pprof (stopped) | Memory | 28.9 KB + 16 B/page | 0.9 KB stubs |
| | CPU | +3.4 instr/pair (+10% cycles) | 0 |
| memory-events | Memory | 4.0 KB | ~0.3 KB stubs |
| | CPU | **+9 to +13 instr/pair (18–26%)** | 0 |
| DHAT (stopped) | Memory | 9.8 KB | 0.5 KB stubs |
| | CPU | 0 (byte-identical) | 0 |
| heap snapshot | Memory | 21.9 KB | ~0.1 KB stubs |
| heap dump | Memory | 3.8 KB | ~0.1 KB stubs |
| owner-gate | Memory | 16.4 KB | 0 (already off) |
| | CPU | +95 instr/pair (+206% cycles) | 0 |

Allocator behavior stays compiled in unconditionally: scavenger, page-holes, purge-all, zero-tracking, fork handlers — all 0 instr/pair on malloc/free (Bun parity).

## Per surface

**C core** (`29001f14`)
- `internal.h`: the inline memevt hook sites are guarded by `#if MI_MEMEVT || MI_DHAT` — DHAT dispatches through the memory-events slow paths, so `MI_MEMEVT=0 MI_DHAT=1` must keep them. `MI_OBSERVERS_MEMEVT_UNRESOLVED` drops out of `MI_OBSERVERS_INITIAL` when memory-events is out, exactly as DHAT's already did, or the fast path takes the slow branch forever. With neither observer compiled in, every hook site expands to nothing and the call sites in `alloc.c`/`free.c`/`alloc-aligned.c` are untouched.
- `mi_unwrapped_malloc/_free/_realloc` and `_mi_memevt_suppress_begin/_end` stay **real** in every build: raw-OS and per-thread-counter helpers, unrelated to tracking.
- `heap-snapshot.c` / `heap-dump.c` (and `arena.c`'s include of `diagnostic-walk.c`, whose only consumer is the dump) compile away under `MI_DIAGNOSTICS`. `mi_option_snapshot_on_exit` stays in every build (Bun parity contract).
- `mi_page_t::metadata` / `has_metadata` now sit behind `#if MI_PPROF` (page header 192 → 176 B with the profiler out). `init.c`'s empty-page sentinel follows; `free.c`/`page.c`/`arena.c` were already guarded.
- CMake registers the memory-events and diagnostics tests only where those subsystems are built. `test-observer-scaling` deliberately stays unconditional — it measures the allocator's scaling, which must hold in the minimal build too, where it is the positive control that compiling the hooks out did not break the fast path. `test-dhat`'s composition half branches on `#if MI_MEMEVT` and asserts the stubs otherwise.

**Rust** (`1916c217`)
- `build.rs` maps all five features to their defines, always 0 or 1.
- `[[test]]` `required-features`: `t18_heap_snapshot` → `diagnostics`; `t20_memory_events` / `t21_visit_live` → `memory-events` (they had no entry at all before). `feature_contract.rs` now asserts **both halves** of all five features. `lib.rs`'s heap-dump unit test gained an inert counterpart.
- `stress-harness`'s dev-dependency names `features = ["pprof"]`: its `allocator_scenario` test drives the profiler, and with `default = []` a workspace-wide `cargo test` would have failed on it.
- Version 0.12.0 + CHANGELOG entry stating the breaking default change and `features = ["full"]` as the restore path.

**CI** (`f8d53dcf`) — every lane that relied on a default now names the flag. New `memevt-only` row; `dhat-on` stays `MI_MEMEVT=0 MI_DHAT=1` (the guard's load-bearing shape) while `debug-full` covers the both-on pair; `pprof-off` becomes the minimal row with every switch spelled out and asserted; `gated` gains `MI_DIAGNOSTICS=ON` because #374's ownership protocol exists for that build. Bundles, ASan, the native `cl` trees and the release asset follow. The cargo lanes in `cross.yml`/`macos-bundles.yml`/`windows-bundles.yml` name the features they used to get implicitly — cargo *skips* a target whose `required-features` are unmet rather than failing, which is exactly how this change could have silently deleted cross-platform coverage. `rust-native.yml` gains a lane per feature plus `--features full`, and clippy covers each compiled-out combination. Features are tested individually and all-on; no cross product (#307).

**New gate** — `ci/check_fastpath_identity.py` gained a third comparison: the **minimal build must be byte-identical to upstream microsoft/mimalloc at the pinned v3 overlay base**. It also pins `-DMI_MEMEVT=ON` alongside `-DMI_DHAT=ON` on the base/HEAD comparison, for the same reason #411 pinned DHAT: otherwise the default flip skews it.

**Docs** — CLAUDE.md rules 3 and 6, `docs/ci-gates.md`, `docs/c-integration.md`, `docs/dhat-and-memory-events.md`, both READMEs. The four claims #414 lists as wrong are corrected: owner gate +206% not "a few percent"; memory-events 9–13 instr/pair when idle, not "one relaxed flag check and nothing else"; `mi_option_t` is 62 enumerators / 14 from this fork at 47–60, not 61/13/47–59; `purge_zeroes` is live again (#337), not dead since #80.

## Evidence (local, gcc 15.2)

- **`ctest`**: minimal **42/42**, everything-on **58/58**, `MI_MEMEVT=0 MI_DHAT=1` **50/50**, `MI_MEMEVT=1 MI_PPROF=0` **45/45**.
- **`ci/check_fastpath_identity.py --base origin/main`** — all four comparisons PASS:
  ```
  PASS: no lock-prefixed instruction in the default build's fast path
  PASS: the minimal build's fast path is byte-identical to upstream 6def7be9458f
  PASS: the default build's fast path is byte-identical to the base revision
  PASS: positive control -- the gate changes mi_malloc, mi_zalloc, mi_free, mi_heap_malloc_small, mi_malloc_small
  ```
  `mi_malloc` (23 instructions), `mi_zalloc` (47), `mi_free` (51), `mi_malloc_small` (21), `mi_heap_malloc_small` (29) — all identical to upstream in the minimal build.
- **Cargo**: `cargo test -p mimalloc-pprof` green with no features, with each of `pprof`/`memory-events`/`diagnostics`/`dhat`/`owner-gate` alone, and with `--features full`. `cargo fmt --all --check` clean. `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean, and likewise for every `--no-default-features [--features X]` combination.
- **Python**: `ruff check` / `ruff format --check` / `pyright` clean; `pytest ci/tests` **472 passed** (includes the verify_local and bundle-table drift guards); `verify_local.py --selftest` PASS (16 configs, 16 bundles); `check_doc_snippets.py` PASS; `check_bun_surface.py` PASS; `check_rust_surface.py` PASS (40 fork exports bound, 62 `mi_option_t` enumerators mirrored in order).

## Judgement calls, stated rather than hidden

- **`pprof-off` was not renamed to `minimal`.** The name is load-bearing in the bundle/manifest plumbing, the coverage comparison and the drift guards; the row now *is* the minimal build (every switch explicit, `MI_MEMEVT=0` asserted) and says so in a comment. Say the word and it becomes `minimal`.
- **`azure-pipelines.yml` and `.github/workflows/test.yaml` were left alone.** Both are upstream-inherited matrices with no `-DMI_PPROF` anywhere, neither is a PR gate, and what they now build — the plain allocator — is exactly what they were written to test. Adding fork flags there would only make upstream cherry-picks dirtier (rule 2).
- **Cross/bundle cargo lanes use `pprof,memory-events,diagnostics`, not `full`.** That is *exactly* what they built before the flip; `owner-gate` changes allocator behaviour and already has rows of its own.

Closes #414. Follow-up: **#415** — folding the observer check into the branch `mi_free` already takes, cold-outlining the slow paths, and the `mi_decl_hidden` one-liner would make the memory-events hooks free, at which point that one could default back on at no cost.


zackees:
Two CI findings on the first two pushes, both now addressed:

1. **`ctest (windows-latest)` (owner-gate tree) and `address sanitizer (clang, Debug, owner-gate)`** — `test-diagnostic-walks` asserted on `mi_memory_set_callbacks`, but it is registered under `MI_DIAGNOSTICS`, which is independent of `MI_MEMEVT`; those two rows deliberately build the dump *without* memory-events so both shapes get compiled. Fixed in `dffcef0d`: the capture is asserted in every such build, and the "no hooked buffer resize" control runs wherever memory-events is compiled in, asserting the stubs otherwise (same shape as test-dhat's composition half). Reproduced locally in the exact failing configuration — 2 failures before, both pass after.

2. **`test-no-pprof` / `Test the everything-on build`** — an allocator debug assertion, `_mi_page_purge_holes_begin: "tld != NULL && !tld->holes_sweeping"`, when `purge_all` runs concurrently with ordinary allocation in an `owner-gate` build. **This is not from this PR**: I reproduced it on `origin/main` (`4aee27a1`) with the same four concurrent sections under `--features dhat,owner-gate` — **2 aborts in 80 runs**, and **0 in 80** without `owner-gate`. Filed with the full reproducer as **#417** rather than papered over.

   What this PR does about it (`bc318c5a`) is make `tests/feature_contract.rs` a single serialized `#[test]` instead of four concurrent ones. That is the convention `t20_memory_events.rs` and `t21_visit_live.rs` already document for process-global state, and it was needed anyway: all four sections flip process-global flags (profiler on/off, DHAT on/off, tracking on/off), so running them on cargo's four harness threads was racing on the very flags they assert. Measured after: **0 aborts in 120 runs** with `--features full`, and the test passes with no features, each feature alone, and `full`.

